PR #1, opened 2022-07-17
Add `How to add a custom function to Context`

bytemain:


yusukebe:
Hi @bytemain ! Thank you for your PR.

It's a nice idea, but we can't recommend extending `Context` by the end-user because it's a little bit tricky.

Although we can implement `c.error`,  there is a `app.onError` method. In `c.message` case, we may not have to make the method in Context.

PS.
The pull request for the documentation is very helpful.

bytemain:
Thanks for the reply, I have read the koa document:
CleanShot 2022-07-18 at 16 02 55@2x

I hope there will have a official way to do `Extend Context`.

Like: https://www.eggjs.org/basics/extend#context

PR #8, opened 2022-08-29
fix typo in cache middleware documentation

NOBLES5E:


yusukebe:
Hi @NOBLES5E !

Thank you, but it's not a typo.

Middleware also works with the `get` method. e.g., the following application using Cache Middleware with `get` method is working.

https://github.com/yusukebe/r2-image-worker

PR #9, opened 2022-09-10
added docs for cloudflare pages

rishavs:


yusukebe:
Hi @rishavs !

Thank you for the PR.

Sorry, I can not accept this PR. Because I can't recommend the usage of Hono on Cloudflare Pages that you wrote about. For this use case, we should use Cloudflare Workers instead. Hono is not the framework for JAMStack.

If we want to use Hono on Cloudflare Pages, I think it should be used to add a bit of dynamic functionality to a static page project. In any case, we should run the Hono on "Cloudflare Workers".

rishavs:
Not sure I understood the concern. JAMStack need not be static. My own project uses Supabase as a backend and serves dynamic data driven pages.

yusukebe:
One of the concepts of Jamstack is "pre-rendering". Hono does not do "pre-rendering" static files. At least, we cannot recommend Hono as Jamstack framework.

justinnoel:
@rishavs I'm going through your process here for deploying a Hono app to Pages.

My only difference is that I use 

`esbuild --bundle src/index.tsx`

vs

`esbuild --bundle src/index.js`

However, I keep getting this in development:

> $ esbuild --bundle src/index.tsx --format=esm --outfile=dist/_worker.js
  dist/_worker.js  40.0kb
✘ [ERROR] _worker.js is importing from another file. This will throw an error if deployed.
  You should bundle your Worker or remove the import if it is unused.
  
  Do you have any suggestions on overcoming this?
  
  ---
  
  I also noticed this line: https://github.com/honojs/honojs.dev/pull/9/files#diff-dfb818efefe98f65628e489d0d31850a4628533f40664ea0bfdc2f2f61b9098dR105
  
  > | Build command        | `npm run dev` |
  
  Shouldn't it be this instead?
  
  > | Build command        | `npm run build` |

PR #10, opened 2022-09-24
Pass Env as Bindings

morinokami:
`Env` should be passed as `{ Bindings: Env }`, otherwise the following ts error would occur:


yusukebe:
Hi @morinokami !

You are right! Thank you! I'll merge it.

PR #11, opened 2022-09-25
docs(bearer-auth): add description for options, recipe for Fastly Compute@Edge

ThatOneBro:
Finally I was able to find time to start the documentation! I hope to be able to get all of the middleware and some of the things in the main repo documented as well!

I started with the Bearer Auth middleware, and a styling update to fix word wrapping. Please let me know what you think of my changes with the options descriptions, and the recipe section.

Also please let me know how I should format these pull requests in the future, maybe we can make a template if it makes sense 😄

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,9 @@
+{
+  "printWidth": 100,
+  "trailingComma": "es5",
+  "tabWidth": 2,
+  "semi": false,
+  "singleQuote": true,
+  "jsxSingleQuote": true,
+  "endOfLine": "lf"
+}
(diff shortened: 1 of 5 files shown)


yusukebe:
Hi @ThatOneBro !

Wow great!!

Both the options descriptions and the recipe section are good. It is very easy to understand.

> Also please let me know how I should format these pull requests in the future, maybe we can make a template if it makes sense 😄

I thought that we don't need the "template". What exactly do you imagine?

ThatOneBro:
> I thought that we don't need the "template". What exactly do you imagine?
Maybe it's fine without a template, just I wasn't sure! If this is a sufficient PR description, then maybe we don't need one! 😄

yusukebe:
@ThatOneBro 

All right. I'll merge this PR!

PR #12, opened 2022-10-01
docs(cache): cache middleware for Deno, describe cache options

ThatOneBro:
This PR adds mention of Cache middleware support in Deno 1.26+ and adds description of Cache middleware options.


yusukebe:
Hi @ThatOneBro !

Cool! Thanks for the quick response!

PR #14, opened 2022-10-03
docs(basic-auth): describe options, add recipes for multi-user auth

ThatOneBro:
This PR describes the Basic Auth middleware, and also fixes a few mistakes in some other pages, such as removing the deprecated 3rd party validator middleware from the list.

It also updates the types for the Bearer Auth middleware and makes the Bearer Auth `Using on Fastly Compute@Edge` recipe match the improved description from the new Basic Auth page.


yusukebe:
Thanks for covering the missing pieces!

PR #15, opened 2022-10-03
docs(compress): add options for compress middleware

ThatOneBro:
This PR updates the page for the Compress middleware, adding descriptions of the options and clarifying a little about how encodings are selected for a response to the client.


yusukebe:
It's easy to understand. Thanks!

PR #17, opened 2022-10-06
docs(context): update examples of body parsing as of v2.1.0

ThatOneBro:
This PR provides more accurate examples of body parsing as of v2.1.0, as noted in https://github.com/honojs/hono/issues/588

### DIFF
--- a/content/docs/api/context.md
+++ b/content/docs/api/context.md
@@ -60,26 +60,26 @@ app.get('/entry/:id', (c) => {
 })
 
 // Parse Request body of type `multipart/form` or `application/x-www-form-urlencoded`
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @ThatOneBro !

Thank you for writing it. There are the things I would like to see fixed. All of the functions are `async`, so we need to do `await`.

```ts
app.post('/entry', async (c) => {
   const body = await c.req.parseBody()
   ...
 })
```

ThatOneBro:
@yusukebe Good catch! I've correct the errors, so hopefully now this documentation should be accurate!

yusukebe:
Perfect! I'll merge it!

PR #22, opened 2022-10-31
Add environment variables example to basic auth

cd-slash:


content/docs/builtin-middleware/basic-auth.md
```diff
@@ -48,6 +50,27 @@ app.get('/auth/page', (c) => {
...
+
+You can avoid storing the username and password in your code by using environment variables:
+
+```ts
+app.use('/auth/*', async (c, next) => {
+  const auth = basicAuth({
+    username: c.env.BASIC_AUTH_USERNAME,
+    password: c.env.BASIC_AUTH_PASSWORD
+  })
+  return await auth(c, next)
```
yusukebe, line 63: We don't have to `await`.

`return auth(c, next)`

yusukebe:
Hi @cd-slash !

Thank you! It would be better to write it down.

I think it would be better to leave "Usage" as it is and put it in "Recipes" for Cloudflare Workers.

PR #24, opened 2022-11-21
Remove `npm init -y` from "Cloudflare Workers"

### LINKED ISSUE #23: So many dependencies are added to `package.json` after running both `npx wrangler init -y` and `npm init -y`
`npx wrangler init -y` creates `package.json`, so `npm init -y` might not need to be run.

https://github.com/cafedomancer/hono-example/compare/1761f9ef258e23e80f8774a27332e111c67b0284...0b369577d21e131d8ff821f839a1025de9c4b76f

https://github.com/honojs/honojs.dev/blob/a0f900c80f25001acefe7c2a7d7b7dfeebef5e0e/content/docs/getting-started/cloudflare-workers.md#L17-L34

cafedomancer:
because the previous command `npx wrangler init -y` also creates package.json. This pull request closes #23.

### DIFF
--- a/content/docs/getting-started/cloudflare-workers.md
+++ b/content/docs/getting-started/cloudflare-workers.md
@@ -29,7 +29,6 @@ npx wrangler init -y
 Install `hono` from the npm registry.
 
 ```
-npm init -y
 npm i hono
 ```
 


yusukebe:
Thanks!

PR #25, opened 2022-11-26
docs(cache): fix typos

rupampoddar:


yusukebe:
Hi @rupampoddar !

Thank you for this PR.

This is not typos. We can also use `get` for the middleware. Using `get` means "cache only the content by GET request". I think it's good to use `get`.

rupampoddar:
Hi @yusukebe ,

Thank you for the clarification. This is actually very useful in case of the Cache middleware.
I apologize for the inconvenience.

PR #29, opened 2023-01-10
Added route+HTTP method examples for basic auth and bearer auth to restrict only certain operations.

VirtualWolf:
See https://github.com/honojs/hono/discussions/793 for context. :)


yusukebe:
Thank you for your quick response!

PR #30, opened 2023-01-10
fix middleware registration method

hclj37:
from `c.use()` to `app.use()`


yusukebe:
Hi @hclj37 !

You are right! Thanks!

PR #35, opened 2023-02-18
chore: fixing code color contrast on dark mode

mzaien:
Hello 👋 congrats on V3 launch 

This or fixes a small thing, as you can see below color contrast for `code`  blocks on dark mode 

CleanShot 2023-02-18 at 19 43 19@2x

And here is the final look after adding the fix

CleanShot 2023-02-18 at 19 38 43


yusukebe:
Hi @mzaien !

Good point! I'll merge it. Thanks!

PR #41, opened 2023-02-24
fix: typo where RPC was PRC

kosei28:


yusukebe:
That's right! Thanks.

PR #43, opened 2023-02-26
fix: Use `bind` with custom fetch method

JonasHiltl:
When passing the custom fetch method for a Cloudflare Service binding, it is necessary to use `bind` since without it an `Illegal Innvocation` Error is thrown.


yusukebe:
Hi @JonasHiltl !

Thank!

PR #45, opened 2023-03-03
Update: 4th step path to index.ts

filipe-freire:
Unless the user remembers to cd into the `src` folder, executing the 4th command in the instructions will result in a `error: missing script "index.ts"` since the user is in the wrong path.

Minor fix, but makes the experience for first timers a little bit more straightforward & less error-prone :)

### DIFF
--- a/getting-started/bun.md
+++ b/getting-started/bun.md
@@ -43,7 +43,7 @@ export default {
 Run the command.
 
 ```ts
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @filipe-freire !

Good fix! I’ll merge it. Thanks!

PR #46, opened 2023-03-13
Changed Interface to Type

spenceradolph:
It appears the above example did not work until changing the interface for Bindings into a Type instead.


spenceradolph:
This example is what I followed to make the changes. It was unclear to me if there’s a distinction between using an interface vs type. 
https://hono.dev/getting-started/cloudflare-workers#bindings

yusukebe:
Hi @spenceradolph !

You are right! We have to use `type` for Bindings. Thanks!

PR #47, opened 2023-03-26
fix duplicated backquote

makies:
It may be a typo.


yusukebe:
Hi @makies !

Yes, this is a typo! Thanks!

PR #48, opened 2023-03-29
Update cloudflare-pages.md

ken0x0a:
for v4.

(I saw deprecation notice)


yusukebe:
Hi @ken0x0a !

Thanks. I recommend the following approach. What do you think?

```ts
const app = new Hono().basePath('/api')
```

ken0x0a:
Oh, I didn't expect that work... it worked... interesting.
I thought that works like [grouping](https://hono.dev/api/routing#grouping) and acts like sub-route (`book` in document URL).

I updated! (force pushed).

yusukebe:
Hi @ken0x0a !

I am sorry for the delay! Thanks. I'll merge it.

PR #53, opened 2023-05-11
Replace npm: to deno.land/x/hono

ryuapp:
I think it's better to use `deno.land/x/hono` on Deno because `npm:` can't move on Deno Deploy now.
But is `deno.land/x/hono` unstable?

- Change `npm:` to `deno.land/x/hono`
- add missing codes

### DIFF
--- a/getting-started/deno.md
+++ b/getting-started/deno.md
@@ -21,13 +21,17 @@ npm create hono@latest my-app
 
 Move into `my-app`. For Deno, you don't have to install Hono explicitly.
 
+```
+cd my-app
+```
+
 ## 3. Hello World
 
 Write your first application.
 
 ```ts
+import { Hono } from 'https://deno.land/x/hono/mod.ts'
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @ryuapp !

Thank you for creating the PR.

You are correct! I've been considering this. It's stable, and we need to use `deno.land/x` to run on Deno Deploy.

The reason I used `npm:` was because it caused a type mismatch with other 3rd party middleware such as Zod Validator. These are managed in the mono repo "**honojs/middleware**", so we currently don't distribute them from `deno.land/x` as we would have to link `deno.land/x` to the GitHub project. To use Zod Validator, we have to import it using `npm:@hono/zod-validator`. However, if we do this, both `deno.land/x` and `npm:` would be in the same script, causing a type mismatch.

But, I have some ideas to resolve this problem. Moreover, enabling use on Deno Deploy is more important than fixing this problem. It's better to use `deno.land/x`.

I'll merge this now!

PR #63, opened 2023-06-22
Remove outdated Fastly Compute@Edge Recipe

JakeChampion:
Fastly Compute@Edge JS supports the SubtleCrypto API, which means the [default built-in sha256 function in hono](https://github.com/honojs/hono/blob/387d6963036a750e5c5dd0a6776a8b67891a659a/src/utils/buffer.ts#LL30C24-L30C24) works :-D


yusukebe:
Hi @JakeChampion !

Ah, you are right! I know Compute@Edge supports that but, it's not updated. Thanks!

PR #73, opened 2023-07-22
Opt for "bun create" in the bun guide

ap-1:
Use [`bun create`](https://bun.sh/docs/templates#bun-create) with the [hono template](https://github.com/bun-community/create-templates/tree/main/hono), because most people creating a project using bun will opt for the `bun` package manager over `npm`


yusukebe:
Hi @ap-1 !

Thanks. Let's go with it!

PR #80, opened 2023-08-04
docs: add valibot validator to 3rd-party middleware page

ZerNico:
Valibot should also be added to the list.

Thanks!


yusukebe:
You are welcome!! Merge now.

PR #88, opened 2023-08-31
docs: file upload/router example

krsbx:
add example for the router and file upload

### DIFF
--- a/guides/file-upload.md
+++ b/guides/file-upload.md
@@ -1,0 +1,42 @@
+# File Upload
+
+How about file upload? Is it possible?
+
+## Basic File Upload
+
+```ts
+import { Hono } from 'hono'
+
+const app = new Hono()
+
+app.post('/upload', async (c) => {
+  const { image } = await c.req.parseBody()
+
+  if (image instanceof File) {
+    const buffer = await image.arrayBuffer()
+  }
+
+  return c.text('Uploaded!')
+})
+```
+
+## File Upload and Save to Disk
+
+```ts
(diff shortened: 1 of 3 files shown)


vinay-osseed (approved):

guides/file-upload.md
```diff
@@ -0,0 +1,42 @@
...
+const app = new Hono()
+
+app.post('/upload', async (c) => {
+  const { image } = await c.req.parseBody()
+
+  if (image instanceof File) {
+    const arrayBuffer = await image.arrayBuffer()
+    const buffer = Buffer.from(arrayBuffer)
+
+    fs.writeFile(path.resolve(__dirname, image.name), buffer)
```
NuroDev, line 37: An import needs to be added for this line, no?

krsbx: I don't think it's necessary to add one since it's an example and let the developer who use `hono` figuring out how they need to adapt the code to their deployment environment.

yusukebe:
Hi @krsbx 

Sorry I didn't reply.

On the router page, I think it's not necessary to describe the usage on the concept page. I want to show only the concept on that page. And we have presets `hono/tiny` and `hono/quick.` so most people will use it. It's not necessary to use the router directly.

About uploading files. We now have the pages below, though it should be improved:

https://hono.dev/snippets/file-upload

krsbx:
@yusukebe i'll make a new PR to improve the page instead

yusukebe:
@krsbx Thanks!

PR #92, opened 2023-09-09
Fix: replace app.get with app.use in Secure Headers middleware

MonsterDeveloper:
Currently the [Secure Headers page](https://hono.dev/middleware/builtin/secure-headers) contains some code examples that use `app.get` instead of correct method for the middleware — `.use`.

This PR replaces all `.get("*")` misuses with `.use("*")`.


yusukebe:
@MonsterDeveloper 

Hi! You are right. It should be `app.use()`. Thanks!

PR #95, opened 2023-09-11
docs: Add usage of jwt with payload in hono/jwt middleware

sor4chi:
Hi, @yusukebe 
I added this because when using the hono/jwt middleware, I didn't realize that payload was available until I went to the source code.
If you like it, I hope you'll MERGE it.

---

BTW, when using jwt on the edge, the only library I could find was https://github.com/tsndr/cloudflare-worker-jwt. are there any plans to release utils/jwt as a package like @hono/jwt?

### DIFF
--- a/middleware/builtin/jwt.md
+++ b/middleware/builtin/jwt.md
@@ -36,6 +36,24 @@ app.get('/auth/page', (c) => {
 })
 ```
 
+Get payload:
+
+```js
+const app = new Hono()
+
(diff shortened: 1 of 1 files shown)


sor4chi:
I understood, thank you!

yusukebe:
@sor4chi 

This issue is created now. Check it out!
https://github.com/honojs/hono/issues/1443

PR #104, opened 2023-09-23
docs: Add deno deploy update

### LINKED ISSUE #100: npm:Hono now supported on deno deploy
I think we should now only recommend npm:Hono in the [getting started tutorial for deno](https://hono.dev/getting-started/deno)

NicoPlyley:
Added information about the update to deno deploy supporting npm.

closes #100


yusukebe:
You are right, `npm:` is not well tested with Deno Deploy. It is also said that using `npm:` reduces performance. Anyway, we need to investigate, but for now, this fix is good.

Thanks!

PR #105, opened 2023-09-24
docs: Add all helper links

NicoPlyley:
In the helper guide there is a link to all the helpers, however, a few were missing due to recent updates


yusukebe:
Thanks!

PR #106, opened 2023-09-25
docs: Update note for bun users

NicoPlyley:
Added note how this middleware is not available in bun since it does not yet support `CompressionStream`


yusukebe:
@NicoPlyley 

Ah, exactly right. thanks!

PR #108, opened 2023-09-26
docs(helper): streaming-sse

watany-dev:
I've added detailed documentation for the "streaming Helper" which provides a method to extend c.stream. 

Ref: https://github.com/honojs/hono/pull/1504


NicoPlyley:
@watany-dev I have a recommendation to add a link to /api/context#stream where you mention is extends `c.stream()`

Also in the guides/helpers can you add the link to this as well.

watany-dev:
@NicoPlyley 
thx!

PR #109, opened 2023-09-28
changed example for /api/routing#including-slashes

vanodevium:
This variant is more correct for routes which ends `.png`


yusukebe:
@vanodevium 

Thanks!

vanodevium:
@yusukebe 

Thank you for incredible framework!

PR #118, opened 2023-10-07
Update validation.md

leocabeza:
Add param as an option to the validator function


yusukebe:
Ah, missing it. Thanks!

PR #120, opened 2023-10-13
feat(validation.md): add info about using multiple validators

MonsterDeveloper:
Currently, the documentation doesn't mention anything about using multiple validators in the same request handler. Yet I've just found out that one can include multiple validators for validating different parts of a request. Moreover, typesafety will also work:

```ts
app.get("/:id",
  zValidator("param", z.object({ id: z.string() })),
  zValidator("json", z.object({ name: z.string() })),
  //...
```


yusukebe:
Thanks! Merging now.

PR #131, opened 2023-10-23
Update jwt.md

CovertCode:
methods: sign, decode, verify are not available on 'hono/jwt' They are available at "hono/utils/jwt/jwt"


yusukebe:
Hi @CovertCode 

No, you can use them from `hono/jwt`.

yusukebe:
This issue should be closed.

PR #139, opened 2023-10-30
docs: Fixes code in AWS Lambda@Edge callback example

talhabalaj:
Just a simple duplication of `from` statement fix.


yusukebe:
Thanks!

PR #140, opened 2023-11-02
Fix Deno's links

ryuapp:
`deno.land` redirect to `deno.com` in basically.
So I changed `deno.land` to `deno.com`.
And I changed the links to the correct docs.


yusukebe:
Thanks!

PR #146, opened 2023-11-06
Lambda Adaptor: streaming response

watany-dev:
I would appreciate it if you could enable this feature after the ticket has been merged.
https://github.com/honojs/hono/pull/1625


yusukebe:
Thanks! Now released v3.10. Merging!

PR #165, opened 2023-11-19
Update context.md

tirsop:
Keep the expression consistent with the rest of the expressions in this documentation page, where we have the explicit return.


yusukebe:
Okay! I'll merge this now. Thanks!

PR #173, opened 2023-12-05
docs: improve the auth bearer multiple token example

mstibbard:
- Example now uses `app.on` to restrict by HTTP Methods
- Example route is now consistent with prior examples (`/api/page`)


yusukebe:
@mstibbard 

Good example! Thanks!

PR #176, opened 2023-12-13
Update vercel.md

mkv27:
Isn't this mandatory to avoid this [issue](https://github.com/honojs/node-server/issues/84) when using with NextJS?

### DIFF
--- a/getting-started/vercel.md
+++ b/getting-started/vercel.md
@@ -148,7 +148,11 @@ Next, you can utilize the `handle` function imported from `@hono/node-server/ver
 import { Hono } from 'hono'
 import { handle } from '@hono/node-server/vercel'
 
-
+export const config = {
+  api: {
+    bodyParser: false,
+  },
+}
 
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @mkv27 

> Isn't this mandatory to avoid this [issue](https://github.com/honojs/node-server/issues/84) when using with NextJS?

No. To avoid the issue, you have to add `NODEJS_HELPERS=0` to your project.

mkv27:
Hello! But what if you need those helpers in other parts of your project?

mkv27:
I just noted there is a similar issue with two different sources, I mean the issue is the same "500 error when the method is not GET or HEAD".
- When you have a node project with the api folder and would like to deploy it in Vercel you'll need the `NODEJS_HELPERS=0`
- When you have a NextJS project and would like to deploy it in Vercel you just need:
```ts
export const config = {
	api: {
		bodyParser: false,
	},
}
```

PR #189, opened 2024-01-03
Remove Netlify from the others list as it has an existing page

Brenden2008:
Netlify already has a page on the website, this commit removes it from the others list.


yusukebe:
Hi @Brenden2008 

Thanks!

PR #192, opened 2024-01-07
add `c.body()`

ryuapp:
I add description of `c.body()`.
And I move the common response description in `c.text()` to `c.body()`.

### DIFF
--- a/api/context.md
+++ b/api/context.md
@@ -41,13 +41,11 @@ new Response('Thank you for coming', {
 })
 ```
 
-## text()
-
-Render text as `Content-Type:text/plain`.
-
+## body()
+Return the HTTP response.
 ```ts
-app.get('/say', (c) => {
-  return c.text('Hello!')
+app.get('/welcome', (c) => {
+  return c.body('Thank you for coming')
 })
 ```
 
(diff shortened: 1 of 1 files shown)


api/context.md
```diff
@@ -41,26 +41,35 @@ new Response('Thank you for coming', {
...
-
-Render text as `Content-Type:text/plain`.
-
+## body()
+Return the HTTP response.
 ```ts
-app.get('/say', (c) => {
-  return c.text('Hello!')
+app.get('/welcome', (c) => {
+  return c.body('Thank you for coming')
```
yusukebe, line 48: Though we can specify it with `c.header,` when you use `c.body,` we should add `Content-Type.` For text, `text/plain` is automatically attached without `Content-Type`, but it is not recommended.

yusukebe:
Hi @ryuapp 

Adding `c.body()` is good, but I made one comment.

ryuapp:
Thank you for the comment.
I removed _Shortcuts for the Response_ section and add `c.body()` description more.
There is some overlap in these descriptions, so I think the description about `c.body()` is sufficient.

I think the shortcuts section should be on another page like a cheet sheet page.

yusukebe:
@ryuapp 

Thanks!

But in addition to this, I would like to say two things.

1. `c.body()` can take a second and third argument.

```ts
app.get('/welcome', (c) => {
  return c.body('Thank you for coming', 201, {
    'X-Message': 'Hello!',
    'Content-Type': 'text/plain',
  })
})
```

2. When returning Text or HTML, it is recommended to use `c.text()` or `c.html()`.

yusukebe:
@ryuapp 

Thanks a lot! If you have another opportunity, please contribute!

PR #193, opened 2024-01-07
Add env section to testing guide

drwpow:
## Suggestion

I had a little bit of trouble getting a good testing setup for Hono + Cloudflare Workers, because they use so many `env` vars. I read this testing guide initially, but as it gave no hints to env, I tried a bunch of different solutions (even including some complicated Vitest code replacement) before I realized this was so easy to pull off.

Open to feedback/suggestions if there’s a better way to do this (i.e. if there’s a better recommended way), but I found this a relatively simple, flexible setup for my own testing.

### DIFF
--- a/guides/testing.md
+++ b/guides/testing.md
@@ -74,3 +74,18 @@ test('POST /posts', async () => {
 ```
 
 In this way, you can test it as like an End-to-End.
+
+## Env
+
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @drwpow 

This is a good suggestion.

For mocking, this approach is great. I'll merge this for it.

And, another approach, we can use [Wrangler's `unstable_dev`](https://developers.cloudflare.com/workers/wrangler/api/#unstable_dev), but it's not documented. The issue: #194

PR #196, opened 2024-01-09
Fix next.js pages router example

Armster15:
Noticed the example for the Next.js pages router used the app router syntax.

1. For edge functions in the page router, export a `config` variable, not `runtime` (https://vercel.com/docs/functions/edge-functions/edge-functions-api)
2. Route handlers are not supported in the pages router


yusukebe:
@Armster15 

You are right! Thanks!

PR #199, opened 2024-01-11
fix text and background are the same color in tip block section

### LINKED ISSUE #197: tip block section in 'code' text color same match
While I was browsing 'JWT Auth middleware' documentation, I found that tip block section there was **'code**' type of text color is same as background color of tip block. it was confused what was written. Kindly try to add white background color of **'code'** type text  and keep same text color as it is.

image

ryuapp:
fix #197 
I add `--vp-c-orange-soft`(rgba(227, 96, 2, 0.1)) as theme color with high transparency.
And I set that to `--vp-c-brand-soft`.

image
image

https://hono.dev/middleware/builtin/jwt

### DIFF
--- a/.vitepress/theme/custom.css
+++ b/.vitepress/theme/custom.css
@@ -13,6 +13,7 @@
   --vp-c-text-dark-3: rgba(235, 235, 245, 0.38);
 
   --vp-c-orange: #e36002;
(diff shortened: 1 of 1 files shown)


yusukebe:
@ryuapp 

Cool! Thanks!

PR #200, opened 2024-01-15
docs: add `Can't modify immutable headers.` tips to proxy example

sor4chi:
https://github.com/honojs/hono/issues/1572

スクリーンショット 2024-01-15 午前11 54 43

### DIFF
--- a/guides/examples.md
+++ b/guides/examples.md
@@ -90,7 +90,7 @@ app.get('/', async (_c) => {
 })
 ```
 
-Since `fetch()` returns unalterable headers, any subsequent attempt to rewrite them with `cors` middleware or the like will result in an error.
+The headers of `Response` returned by `fetch` are immutable. So, an error will occur if you modify it.
 :::
 
 ## JSX
--- a/guides/examples.md
+++ b/guides/examples.md
@@ -78,6 +78,21 @@ app.get('*', (c) => {
 export default app
 ```
 
+::: tip
+If you can see `Can't modify immutable headers.` error with similar code, you need to clone the response object.
+
+```ts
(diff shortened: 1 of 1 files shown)


guides/examples.md
```diff
@@ -78,6 +78,21 @@ app.get('*', (c) => {
...
+```ts
+app.get('/', async (_c) => {
+  const response = await fetch('https://example.com')
+  // clone the response to return a response with modifiable headers
+  const newResponse = new Response(response.body, response)
+  return newResponse
+})
+```
+
+Since `fetch()` returns unalterable headers, any subsequent attempt to rewrite them with `cors` middleware or the like will result in an error.
```
yusukebe, line 93: Rewrite like the following:

"The headers of `Response` returned by `fetch` are immutable. So, an error will occur if you modify it."

yusukebe:
Hi @sor4chi 

Please check the comment.

sor4chi:
@yusukebe fixed!

guides/examples.md
```diff
@@ -78,6 +78,21 @@ app.get('*', (c) => {
...
+```ts
+app.get('/', async (_c) => {
+  const response = await fetch('https://example.com')
+  // clone the response to return a response with modifiable headers
+  const newResponse = new Response(response.body, response)
+  return newResponse
+})
+```
+
+The headers of Response returned by fetch are immutable. So, an error will occur if you modify it.
```
yusukebe, line 93: Please make `Response` and `fetch` as code spans.

sor4chi: opps, sorry!

yusukebe (approved):
LGTM!

PR #201, opened 2024-01-18
docs: update custom context renderer type

sor4chi:
スクリーンショット 2024-01-18 11 57 09
Hi, @yusukebe 
The type that was in the documentation could not support jsx that returned `Promise<string>`, so a type was added.


yusukebe:
@sor4chi 

Thanks!

PR #203, opened 2024-01-23
docs: update typo and add warning to `$url()` to use absolute URL

kndwin:
Happy to change the wording, I had to dig around the debugging breakpoints to figure this out so thought it might be helpful to make it more clear.

image

### DIFF
--- a/guides/rpc.md
+++ b/guides/rpc.md
@@ -136,7 +136,23 @@ const client = hc<AppType>('/api', {
 
 ## `$url()`
 
-You can get a `URL` object for accessing the the endpoint by using `$url()`.
+You can get a `URL` object for accessing the endpoint by using `$url()`.
+::: warning
+You have to pass in an absolute URL for this to work. Passing in a relative URL `/` will result in the following error.
+
+`Uncaught TypeError: Failed to construct 'URL': Invalid URL`
+
+```ts
+// ❌ Will throw error
+const client = hc<typeof route>('/')
+client.api.post.$url()
+
+// ✅ Will work as expected
+const client = hc<AppType>('http://localhost:8787')
+client.api.post.$url()
+```
+
+:::
 
 ```ts
 const route = app


yusukebe:
Hi @kndwin !

This is a nice improvement. However, maybe if you run this in a browser, it will work with relative URLs. It is a wrapper for `fetch`, so it behaves the same as `fetch`.

So why not say "If you get an error, specify an absolute URL"?

kndwin:
I actually tried it in a browser, the annoying(?) thing is `fetch("/api")` works well, but when you use the `new URL()` class it breaks.

## Investigation
Simple server code just to get the response
image

My server is returning the correct response
 
```
GET /api/auth/echo 200 OK (7.71ms)
GET /api/auth/echo 200 OK (8.99ms)
GET /api/auth/echo 200 OK (4.23ms)
GET /api/auth/echo 200 OK (3.20ms)
```

This is the browser that is served from the server, what's interesting is the comparison between `fetch()` and `new URL()` with relative vs absolute paths

image

`new URL()` is the culprit and I suspect is the below code here
https://github.com/honojs/hono/blob/main/src/client/client.ts#L144-L149

Looking through the debugger with breakpoints, we can see it aligns with we expect

image
image

image

## Possible next steps
1. Short term fix is to just update the documentation, let me know if the above assumption that it works in the browser still holds true. I might be missing something from my investigation and happy to update it with your suggestion. If not, I'm still very flexible with the wording (happy to do a few drafts)

3. Long term fix could be to do a PR to handle this error (Happy to give it a go if you are okay with that)
There's a few angles I can think of, I'll list them out but 
   -. Force absolute URL at the type level
   -. Let the error throw and return a more descriptive error message
   -. Somehow get the `baseUrl` and pass it into `new URL()`

yusukebe:
@kndwin 

Super sorry! You are right. It should be an absolute URL.

I'll merge this PR now.

> Let the error throw and return a more descriptive error message

I think it is a simple way to throw an error while other methods may complicate the code. If possible, I would like you to make a PR.

Thanks!

PR #218, opened 2024-02-01
docs: Mention nesting selector in css documentation

tmcw:
It wasn't immediately clear how to style hover effects using Hono's `css` helper, but it's very possible - this adds the example to the documentation.


yusukebe:
@tmcw Good! Merging now. Thanks!

PR #236, opened 2024-02-10
fix bun test example

swdyh:
In the test when using bun, make the request method the same as the non-bun case.

### DIFF
--- a/getting-started/bun.md
+++ b/getting-started/bun.md
@@ -59,6 +59,7 @@ export default app // [!code --]
 export default { // [!code ++]
   port: 3000, // [!code ++]
   fetch: app.fetch, // [!code ++]
+  request: app.request, // [!code ++]
 } // [!code ++]
 ```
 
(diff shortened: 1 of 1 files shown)


getting-started/bun.md
```diff
@@ -59,6 +59,7 @@ export default app // [!code --]
 export default { // [!code ++]
   port: 3000, // [!code ++]
   fetch: app.fetch, // [!code ++]
+  request: app.request, // [!code ++]
```
yusukebe, line 62: This is not necessary.

yusukebe:
@swdyh 

Thanks. I've commented. Check it.

getting-started/bun.md
swdyh: When you say "this is not needed", does that mean you can use app.request without it?
Please tell me why it is not necessary.

Without this line, in my environment(bun 1.0.26, hono 4.0.1, 3.12.12) I get an error like this.

スクリーンショット 2024-02-12 0 42 48

yusukebe: I assume the following case where the port number is not changed.

```ts
const app = new Hono()

// ...

export default app
```

The exported application contains the `request` property.

swdyh: Yes, I understand that if "app" is exported, it has a request property. This proposal is to use "app.request" when "app" is not exported with port change. If you don't need that, then so be it. Thank you.

yusukebe: Thank you for understanding. I want to keep it simple here, so can you please remove that line?

PR #244, opened 2024-02-14
docs: minor typo

missinglink:
docs: minor typo


yusukebe:
Wow, you are right. Thanks!

PR #247, opened 2024-02-16
docs(getting-started): static files on CF Workers

stukennedy:
there is a requirement to add a manifest to the serveStatic adapter method, the docs correctly shows that you need to import a manifest from the special __STATIC_CONTENT_MANIFEST module

However, this results in a Typescript error, as this module is generated by wrangler. This change to the docs explains how the user can add a type definition so that the error goes away, otherwise they may assume that the example is incorrect.

### DIFF
--- a/getting-started/cloudflare-workers.md
+++ b/getting-started/cloudflare-workers.md
@@ -183,6 +183,15 @@ app.get('/static/*', serveStatic({ root: './', manifest }))
 app.get('/favicon.ico', serveStatic({ path: './favicon.ico' }))
 ```
 
+TypeScript cannot resolve the special __STATIC_CONTENT_MANIFEST variable used to import the manifest. This variable is specific to Cloudflare Workers' environment and isn't recognized by TypeScript out of the box.
+You can tell TypeScript about this special module by declaring it in a type definition file. Create a new .d.ts file in your project, such as custom.d.ts, and add the following declaration:
+```ts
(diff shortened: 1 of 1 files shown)


getting-started/cloudflare-workers.md
```diff
@@ -183,6 +183,15 @@ app.get('/static/*', serveStatic({ root: './', manifest }))
 app.get('/favicon.ico', serveStatic({ path: './favicon.ico' }))
 ```
 
+TypeScript cannot resolve the special __STATIC_CONTENT_MANIFEST variable used to import the manifest. This variable is specific to Cloudflare Workers' environment and isn't recognized by TypeScript out of the box.
```
yusukebe, line 186: You can write "__STATIC_CONTENT_MANIFEST" as a code span:

```md
`__STATIC_CONTENT_MANIFEST`
```

getting-started/cloudflare-workers.md
```diff
@@ -183,6 +183,15 @@ app.get('/static/*', serveStatic({ root: './', manifest }))
...
+You can tell TypeScript about this special module by declaring it in a type definition file. Create a new .d.ts file in your project, such as custom.d.ts, and add the following declaration:
```
yusukebe, line 187: It's better to use a code span for ".d.ts" too:

```md
`.d.ts`
```

yusukebe:
Hi @stukennedy 

Thanks! That's a good suggestion. I've left the comments. Please check them.

PR #253, opened 2024-02-23
feat: Support vitepress-twoslash

shinGangan:
### 🙋 Description

Support for using `@shikijs/vitepress-twoslash` in documents.
- refl: https://shiki.style/packages/vitepress

#### 📸 Usage and Captured image

````
```ts twoslash
import { Hono } from 'hono'

const app = new Hono()

app.get('/', (c) => {
  return c.text('Hello Hono!')
})

export default app
```
````

show ↓

### ☑️ Checklist
 
- [ ] linked an issue or discussion
- [x] add example

### 📗 Note

I believe there is still some debate on such issues as "should it be displayed on all pages?"
Therefore, this PR focuses on "building the environment."

This PR is for use when “We want to use Twoslash on Hono.dev as well".

### DIFF
--- a/.vitepress/config.ts
+++ b/.vitepress/config.ts
@@ -1,5 +1,6 @@
 import { defineConfig } from 'vitepress'
 import type { DefaultTheme } from 'vitepress'
+import { transformerTwoslash } from '@shikijs/vitepress-twoslash'
 
 const sidebars = (): DefaultTheme.SidebarItem[] => [
   {
@@ -194,6 +195,7 @@ export default defineConfig({
       light: 'github-light',
       dark: 'github-dark',
     },
+    codeTransformers: [transformerTwoslash()],
   },
   themeConfig: {
     logo: '/images/logo-small.png',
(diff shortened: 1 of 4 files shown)


shinGangan:
Hi @yusukebe .

We consider this PR to be a very low priority for review.
Therefore, we would appreciate a review when you are available. 🙇

yusukebe:
Hi @shinGangan 

This is a super cool feature!!

In Hono, however, we must consider whether to introduce it. This is because the types in Hono are so complex that users often do not immediately understand them when they see them.

For example, the `HandlerInterface` has more than 20 overloads.

Screenshot 2024-02-24 at 22 32 37

So, I am not sure if this is for the benefit of the real users. However, it is an awesome cool feature, so let's leave this PR open.

Thanks!

shinGangan:
> In Hono, however, we must consider whether to introduce it.

I agree with you on this.
The same reason we have not made any updates to the documentation within this PR.

Whether this feature is really necessary for Hono and Hono users needs to be scrutinized within the team.

yusukebe:
@shinGangan 

Thank! Closed. Let's continue in #503 .

PR #258, opened 2024-03-03
docs(top): change in usages

ryuapp:
Polyfill.io server has moved to Rust, so Hono is no longer in use.
Ultra also hasn't been updated for about half a year.

Add OpenStatus instead. I think this is a good usage to use Bun as a platform.

### DIFF
--- a/top.md
+++ b/top.md
@@ -74,7 +74,7 @@ Here are some examples of use-cases.
 | ---------------------------------------------- | ------------------ | ----------------------------------------------------------------------------------------- |
 | [cdnjs](https://cdnjs.com)                     | Cloudflare Workers | A free and open-source CDN service. _Hono is used for the api server_.                    |
 | [Cloudflare Blog](https://blog.cloudflare.com) | Cloudflare Workers | _Some applications featured in the articles use Hono_.                                    |
-| [Unkey](https://unkey.dev/)                    | Cloudflare Workers | An open-source API authentication and authorization. _Hono is used for the api server_.   |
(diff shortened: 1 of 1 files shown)


yusukebe:
@ryuapp 

Good improvement! Agree.

And, how about adding [unkey](https://unkey.dev/)? They are using Hono and Cloudflare heavily.

ryuapp:
@yusukebe 
I agree with your opinion. I added unkey to usages.
In addition, I have also organized the API server representation and organized projects by platform.

yusukebe:
@ryuapp 

Thanks! But I have a few additional things.

I'd like to add "Deno Docs" and "Cloudflare D1".

Deno Docs which is the official document site of Deno is using Hono as a backend server: https://github.com/denoland/deno-docs

Cloudflare D1 team is using Hono to build an "internal" API server for D1.

And, how about adding linkt for this discussion "Who is using Hono in production?": https://github.com/orgs/honojs/discussions/1510

ryuapp:
@yusukebe Awesome, I add them.
By the way, does Cloudflare Blog use Hono? 
I think it's okay to remove it if it's just the articles about apps that use Hono.

yusukebe:
@ryuapp 

> By the way, does Cloudflare Blog use Hono?

That means "examples of Cloudflare Blog posts are using Hono". Like this one:
https://blog.cloudflare.com/making-static-sites-dynamic-with-cloudflare-d1/

Hono is **not** used for the Cloudflare Blog "system".

So, it might make confusion. I think we can remove it.

ryuapp:
@yusukebe 
Thank you for your comment.
I’ve done it all.

yusukebe:
@ryuapp 

Awesome! Let's go with this. Thanks!

PR #259, opened 2024-03-03
fix: type in deno.md

jyotman:
create-hono makes a project with main.ts file but the command in the documentation was instructing to run hello.ts


yusukebe:
@jyotman 

Thanks! That's a typo!

PR #262, opened 2024-03-06
feat: use HonoX

goisaki:
Everyone wants Hono documents to be by Hono.
According to file-based routing by HonoX, it is possible.
It will be interesting to see if this can be an example of a project by HonoX & Hono.
(It should be noted that HonoX is not yet stable.)

> I am ready to accept your PR, so please help me.


goisaki:
@yusukebe, what do you think about this idea?

yusukebe:
Related to https://github.com/honojs/honox/issues/86

yusukebe:
Hi @goisaki 

This is an awesome challenge! I respect you for trying to do this!

Personally, I would prefer to continue using VitePress to lower maintenance costs, but I am eager to support this challenge.

Actually, VitePress very well and quickly transitions pages using the SPA approach. However, even serving HTML is fast enough if distributed from the edge, so this should not be a problem.

We want others' opinions, ideas, and help!

goisaki:
@yusukebe 
I understand. Certainly vitepress would be best for maintenance costs. I will include SPA implementation in my roadmap and try to make it better than vitepress. If you are not negative about this PR I will continue to develop it. Thanks!

goisaki:
@yusukebe 
I would like to reduce the number of filename and content changes in the current document.
(for other PR's that are going on right now).
I also want to avoid changing `.md` to `.mdx` for this reason.
However, HonoX has hard-coded extensions that are omitted in the path, and I don't know of any way to add them here.
https://github.com/honojs/honox/blob/d200fdeddb8bad998c4440dc3975cc1d3f60cb96/src/utils/file.ts#L1-L10
I am considering creating a PR to add this functionality to HonoX, what do you think about this issue?
If this concern is unnecessary please let me know so.

yusukebe:
> I am considering creating a PR to add this functionality to HonoX, what do you think about this issue?

The reason why it is hard-coded is to prevent extensions other than the expected ones. But, allowing `.md` not only `.mdx` makes sense. If you can confirm that the `*.md` works well, you can create the PR to support `.md`!

goisaki:
Hmmm, I've tried locally, but I can't seem to find a simple, minimalist way to do it.
It is not impossible to do that, but it would add a lot of code that is far from the interest of this repository.
For example:

- Syntax highlighting with {language name / copy button}
- Dark mode support
- (SPA)

Also, the addition of the sample project, which is the original purpose, has fortunately already been done in `yusukebe/honox-examples` (https://github.com/yusukebe/honox-examples/tree/main/projects/blog)

So I am beginning to think this PR is unnecessary.
Sorry for the disappointment.
(If there is one other brave challenger who is interested, I will do my best to support them.)
@yusukebe , can you forgive me?

yusukebe:
@goisaki 

No problem!! You are challenge is awesome.

> - Syntax highlighting with {language name / copy button}
> - Dark mode support
> - (SPA)

SPA may be difficult, but I feel it can be done for other things. Also, we can improve HonoX to do that. You may leave this PR if you like. Anyway, what you have done is not a waste. Thank you so much!

PR #271, opened 2024-03-12
Update cloudflare-workers.md

futantan:


getting-started/cloudflare-workers.md
```diff
@@ -11,6 +11,7 @@ Let’s make your first application for Cloudflare Workers with Hono.
 
 A starter for Cloudflare Workers is available.
 Start your project with "create-hono" command.
+Select Cloudflare Workers template for this example
```
yusukebe, line 14: I think it's better to write the actual output name instead of "Cloudflare Workers". Like the following:

```md
Select `cloudflare-workers` for this example.
```

yusukebe:
Hi @futantan 

Thanks for the PR! You are right. We have to write the template name.

This PR includes a fix only for Cloudflare Workers. But we also have to do the same thing for other platforms. So, could you fix the other platform pages? That will be very helpful.

futantan:
hi @yusukebe updated

yusukebe (approved):
LGTM!

yusukebe:
@futantan Great work!

I'll merge this now. Thank you!

PR #272, opened 2024-03-17
updated missing word "not"

Virous77:
"It is suitable for not just only returning JSON."

"not" was missing there. as i feel.

adding "not" strongly  emphasizes that Hono's capabilities extend beyond merely returning JSON.


yusukebe:
Hi @Virous77 

That's right. Thanks!

PR #290, opened 2024-04-05
Add new grouping example

exoego:
When I migrated my app from Express to Hono, I needed this example.

## Express

```typescript
const book = express()
book.get('/book', ...)
book.post('/book', ...)

const user = express()
user.get('/user', ...)
user.post('/user', ...)

const app = user.use(book) // no base path needed
```

## Hono

```typescript
const book = new Hono()
book.get('/book', ...)
book.post('/book', ...)

const user = new Hono()
user.get('/user', ...)
user.post('/user', ...)

const app = user.route('/', book)
```

### DIFF
--- a/api/routing.md
+++ b/api/routing.md
@@ -106,6 +106,24 @@ const app = new Hono()
 app.route('/book', book)
 ```
 
+## Grouping without changing base
+
+You can also group multiple instances while keeping base.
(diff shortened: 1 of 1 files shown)


yusukebe:
Nice! Thanks @exoego

PR #292, opened 2024-04-05
Docs: clarified that RSA JWT uses JWKS

Soviut:
Auth providers like Auth0 prominently feature the term JWKS for their RSA JWT verification so it is good to include the term to make it easier to search for.


yusukebe:
@Soviut 

Looks good to me! Merging now. Thanks!

PR #294, opened 2024-04-06
Properly syntax highlight code blocks

morinokami:
Many command examples were syntax-highlighted as `txt`, so they have been changed to `sh`. Additionally, command examples that were not syntax-highlighted have also been updated to include `sh` highlighting.

| Before | After |
|--------|--------|
| Screenshot 2024-04-07 at 1 22 49 | Screenshot 2024-04-07 at 1 23 07 |


yusukebe:
@morinokami 

You're right. It should be `sh`. I'll merge this PR now.

PR #295, opened 2024-04-07
Update `notFoundOption` to `onNotFound`

morinokami:
`notFoundOption` should be written as `onNotFound`.


yusukebe:
@morinokami Thanks!

PR #296, opened 2024-04-07
Update Bun and Deno run commands

morinokami:
To maintain consistency with other examples that use `npm run dev` as the command for running an app, I've modified the run commands for Bun and Deno. Since the templates for Bun and Deno include the necessary scripts (https://github.com/honojs/starter/blob/main/templates/bun/package.json#L3, https://github.com/honojs/starter/blob/main/templates/deno/deno.json#L3), this should not cause any issues.

### DIFF
--- a/getting-started/bun.md
+++ b/getting-started/bun.md
@@ -41,7 +41,7 @@ export default app
 Run the command.
 
 ```ts
-bun run --hot src/index.ts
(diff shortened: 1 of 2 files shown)


yusukebe:
HHi @morinokami 

Thanks! That's good. Merging now.

PR #297, opened 2024-04-07
Update Vercel page

Armster15:
Hello, and thanks for Hono! I've made two changes to the Vercel docs:

1. I added a TS type to the exported `config` object that some of the API routes use
2. To address https://github.com/honojs/node-server/issues/84, I updated the pages router Node.js example to have the `bodyParser` setting turned off. Based on my manual testing, this issue only affects the pages router API routes running on Node.js: the app router as a whole and page router edge API routes are not affected and as such don't need bodyParser turned off.


yusukebe:
@Armster15 

Thanks! Great. I'll merge this now.

PR #298, opened 2024-04-10
Update routing.md

g3r4n:
In the group exemple we give 2 solutions:
one without basePath and one with basePath.
The second solution seems to not be right in the comment or the implementation. This fix presume the implementation is not right and fix it.


yusukebe:
Hi @g3r4n 

That's right! Thanks!

PR #299, opened 2024-04-10
New question for faq.md

janat08:


janat08:
I actually discovered that there's a integration with sentry that doesn't require code changes, but the error-handling methods could be improved.

PR #300, opened 2024-04-15
export default app to provide working basic jsx example for cf workers

dashedstripes:
The basic JSX example was missing the default export so it would break on cloudflare workers if you copy/pasted that snippet.


yusukebe:
@dashedstripes 

Thanks. That's good!

PR #304, opened 2024-04-17
Remove locale in Fastly link

robvanbakel:
The docs pointed to the Japanese version of the Fastly signup page. This commit updates the link to the generic signup page.


yusukebe:
@robvanbakel 

Yeah, you're right. Thanks!

PR #305, opened 2024-04-18
chore: bump dependency vitepress to stable

PaiJi:
We had a section in our console that reported an error from vitepress, which has been [fixed](https://github.com/vuejs/vitepress/issues/3389), so I think we should update the version, currently vitepress is out of RC to release a stable version.

image

I've tested most of the pages locally and everything looks fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "private": true,
   "license": "MIT",
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


yusukebe:
@PaiJi 

This is good! Thanks.

PR #306, opened 2024-04-19
Format snippets to match code style

mic-sob:
No changes in the content, fixed some spacing and formatting.


yusukebe:
Hi @mic-sob 

Thank you for the PR. I'll merge this now.

I think the best way to unify the format is to create a lint rule and create a method for applying it before merging.

PR #310, opened 2024-04-21
Add language type in code block

tao-s:
For syntax highlight


yusukebe:
Hi @tao-s 

`console`! Honestly, I didn't know we could use `console` as the language for a code block, though I know `sh` or `bash`! There are a lot of part we can use it in our website. If you don't mind, we'd be grateful if you could fix them! `console` will be better than `sh` or `bash`.

yusukebe:
For this PR, looks great. Mergin now. Thanks!

PR #311, opened 2024-04-22
docs: add "cloudflare worker local dev env" section

PaiJi:
I've added how to configure local development variables in cloudflare worker to the documentation as well, please let me know if you have any questions.

@yusukebe

### DIFF
--- a/getting-started/cloudflare-workers.md
+++ b/getting-started/cloudflare-workers.md
@@ -373,4 +373,36 @@ then edit `wrangler.toml`, and add this code after `compatibility_date` line.
 main = "src/index.ts"
 minify = true
 ```
+
 Everything is ready! Now push the code and enjoy it.
+
+## Load env when local development
+
+To configure the environment variables for local development, create the `.dev.vars` file in the root directory of the project.
+Then configure your environment variables as you would with a normal env file.
+
+```
+SECRET_KEY=value
+API_TOKEN=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9
+```
+
+
+> For more details visit the official Cloudflare documentation:
+> https://developers.cloudflare.com/workers/wrangler/configuration/#secrets
+
+Then we use the [env adapter](/helpers/adapter#env) to get the environment variables in our code.
+**For Cloudflare Workers, environment variables must be obtained via `c`, not via `process.env`.**
+
+```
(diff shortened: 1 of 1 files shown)


getting-started/cloudflare-workers.md
```diff
@@ -373,4 +373,36 @@ then edit `wrangler.toml`, and add this code after `compatibility_date` line.
...
+```
+SECRET_KEY=value
+API_TOKEN=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9
+```
+
+
+> For more about this section you can find in the Cloudflare documentation:
+> https://developers.cloudflare.com/workers/wrangler/configuration/#secrets
+
+Then we use the [env adapter](/helpers/adapter#env) to get the environment variables in our code.
```
yusukebe, line 393: We recommend using the `c.env` like a `c.env.SECRET_KEY` instead of the env adapter.

PaiJi: @yusukebe updated🙂

getting-started/cloudflare-workers.md
```diff
@@ -373,4 +373,33 @@ then edit `wrangler.toml`, and add this code after `compatibility_date` line.
...
+
+> For more about this section you can find in the Cloudflare documentation:
+> https://developers.cloudflare.com/workers/wrangler/configuration/#secrets
+
+Then we use the `c.env.*` to get the environment variables in our code.  
+**For Cloudflare Workers, environment variables must be obtained via `c`, not via `process.env`.**
+
+```ts
+app.get('/env', (c) => {
+  const SECRET_KEY = c.env?.SECRET_KEY
```
yusukebe, line 397: `?` is not needed if you specify types correctly:

```ts
const app = new Hono<{
  Bindings: {
    SECRET_KEY: string
  }
}>()
```

We don't recommend to use `?`.

PaiJi: Updated. one curiosity as a beginner though, using it here`?` Would there be any problems with it?

yusukebe: Thanks. It is just a problem, type-safe or not. If you set `Bindings` you don't need it and it will be type-safed.

yusukebe (approved):
LGTM!

yusukebe:
@PaiJi 

Thanks! Merging now.

PR #314, opened 2024-04-25
Update htmx.md to add working instructions for using typed-htmx

ligabloo:
The current example for integrating `typed-html` with `hono/jsx` is outdated and doesn't work on current versions of both packages. I've updated the examples based on the instructions for [augmenting the JSX type definitions](https://github.com/Desdaemon/typed-htmx/blob/main/example/src/types.d.ts) for libraries which inject their own JSX types.

### DIFF
--- a/snippets/htmx.md
+++ b/snippets/htmx.md
@@ -9,15 +9,13 @@ Using Hono with [htmx](https://htmx.org/).
 By using [typed-htmx](https://github.com/Desdaemon/typed-htmx), you can write JSX with TypeScript definitions for htmx attributes.
 We can follow the same pattern found on the [typed-htmx Example Project](https://github.com/Desdaemon/typed-htmx/blob/main/example/src/types.d.ts) to use it with `hono/jsx`.
 
-Install:
+Install the package:
 
 ```sh
 npm i -D typed-htmx
 ```
 
-`app/global.d.ts`:
-
-Import the `typed-htmx` types:
+On `src/global.d.ts` (or `app/global.d.ts` if you're using HonoX), import the `typed-htmx` types:
 
 ```ts
 import "typed-htmx";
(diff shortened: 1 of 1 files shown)


snippets/htmx.md
```diff
@@ -6,22 +6,33 @@ Using Hono with [htmx](https://htmx.org/).
 
 ### typed-htmx
 
-With using [typed-htmx](https://github.com/Desdaemon/typed-htmx), you can write the JSX with the type definitions of htmx.
+By using [typed-htmx](https://github.com/Desdaemon/typed-htmx), you can write JSX with TypeScript definitions for htmx attributes.
```
ligabloo, line 9: I tried to improve wording on this sentence, but I'd be glad to revert it if needed.

snippets/htmx.md
```diff
@@ -6,22 +6,33 @@ Using Hono with [htmx](https://htmx.org/).
...
+We can follow the same pattern found on the [typed-htmx Example Project](https://github.com/Desdaemon/typed-htmx/blob/main/example/src/types.d.ts) to use it with `hono/jsx`.
 
 Install:
 
 ```sh
 npm i -D typed-htmx
 ```
 
-`tsconfig.json`:
+`app/global.d.ts`:
```
yusukebe, line 18: Some users will use `src` and have to make `src/global.d.ts`. So you can add a note about it.

snippets/htmx.md
yusukebe: This is good!

yusukebe:
Hi @ligabloo 

Thank you! I've left a comment, check it.

snippets/htmx.md
ligabloo: Is the `app` folder part of the honox template exclusively?
If so, maybe something like the following?

```md
`src/global.d.ts` (or `app/global.d.ts`, if you're using HonoX):
```

yusukebe: Yeah. Exactly!

ligabloo: Done!

yusukebe:
@ligabloo 

Great! I'll merge this now. Thanks a lot!

PR #316, opened 2024-04-26
docs: fix the type issue of `c.var`

sor4chi:
Sharing the Env types for making enables the `c.var` type-infer whether using the handler with no middleware or not.

https://github.com/honojs/hono/issues/2559

### DIFF
--- a/api/context.md
+++ b/api/context.md
@@ -17,7 +17,7 @@ app.get('/hello', (c) => {
 
 Return the HTTP response.
 
-You can set headers with `c.header()` and set HTTP status code with `c.status`.  
+You can set headers with `c.header()` and set HTTP status code with `c.status`.
 This can also be set in `c.text()`, `c.json()` and so on.
 
 ::: info
@@ -162,13 +162,15 @@ If you want to create the middleware which provides a custom method,
 write like the following:
 
 ```ts
-const app = new Hono()
-
-const echoMiddleware: MiddlewareHandler<{
+interface Env {
   Variables: {
     echo: (str: string) => string
   }
-}> = async (c, next) => {
+}
+
+const app = new Hono<Env>()
+
+const echoMiddleware: MiddlewareHandler<Env> = async (c, next) => {
   c.set('echo', (str) => str)
   await next()
 }
--- a/api/context.md
+++ b/api/context.md
@@ -162,7 +162,7 @@ If you want to create the middleware which provides a custom method,
 write like the following:
 
 ```ts
-interface Env {
+type Env = {
   Variables: {
     echo: (str: string) => string
   }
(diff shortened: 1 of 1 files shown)


api/context.md
```diff
@@ -162,13 +162,15 @@ If you want to create the middleware which provides a custom method,
 write like the following:
 
 ```ts
-const app = new Hono()
-
-const echoMiddleware: MiddlewareHandler<{
+interface Env {
```
yusukebe, line 165: Should be `type`:

```ts
type Env = {
  Variables: {
    echo: (str: string) => string
  }
}
```

api/context.md
```diff
@@ -162,13 +162,15 @@ If you want to create the middleware which provides a custom method,
...
   Variables: {
     echo: (str: string) => string
   }
-}> = async (c, next) => {
+}
+
+const app = new Hono<Env>()
+
+const echoMiddleware: MiddlewareHandler<Env> = async (c, next) => {
```
yusukebe, line 173: I recommend using the `createMiddleware` factory method to avoid defining the types by users them selfs:

```ts
const echoMiddleware = createMiddleware<Env>(async (c, next) => {
  c.set('echo', (str) => str)
  await next()
})
```

yusukebe:
Hi @sor4chi 

I've left comments. Please check them.

sor4chi:
Hi, @yusukebe. I fixed!

api/context.md
```diff
@@ -162,16 +162,18 @@ If you want to create the middleware which provides a custom method,
...
-
-const echoMiddleware: MiddlewareHandler<{
+type Env = {
   Variables: {
     echo: (str: string) => string
   }
-}> = async (c, next) => {
+}
+
+const app = new Hono<Env>()
```
yusukebe, line 171: It does not need passing `Env` as a generics:

```ts
const app = new Hono()
```

sor4chi: As pointed out in the referenced issue, this PR is to solve the problem of type completion when using middleware with use.
As shown in the following image, this type problem can be solved by passing Env to Hono's generics.
Screenshot 2024-04-29 at 19 00 44

Screenshot 2024-04-29 at 19 01 01

sor4chi: Should I write this as a separate section?

yusukebe: I made `createMiddleware()` to make the user not write `Hono<Env>` if added just before the handler. So I don't want to write `Hono<Env>` if it is not necessary.

Then, you can write, `If you want to add your custom middleware separate from handlers, you should pass the `Env` to Hono as generics`, as another example.

sor4chi: Okay, I’ll separate the codes. Thank you

yusukebe: Gooood! Thanks!

PR #319, opened 2024-04-28
Fix mistake in Bun serve static

NicoPlyley:
The reference to `index.ts` made it appear that the static folder was inside the `src` folder when root is set to `./`. Someone opened an issue thinking it was a bug [here](https://github.com/honojs/hono/issues/2565).


yusukebe:
@NicoPlyley 

Thank you for this PR and for handling the issue! That's right. I'll merge this now.

Issue #317, opened 2024-04-26
"How to write your middleware"


yusukebe:
We have to write "How to write your middleware".

* This should include how to override the response after `next`: https://github.com/honojs/hono/issues/960#issuecomment-2078955638
* Recommend use the `createMiddleware()` method.

NicoPlyley:
Where do you want this placed on the website? Is this for snippets or part of the main docs?

yusukebe:
@NicoPlyley 

Hi Nico.

> Is this for snippets or part of the main docs?

The main docs. I think "[/guides/middleware](https://hono.dev/guides/middleware)" will be best.

PR #320, opened 2024-04-29
Added docs for creating middleware

NicoPlyley:
Hi @yusukebe,

I added what was mentioned in #317 to the existing custom middleware section in the guide. I tried to keep it short but informative. I was debating adding another example for adding parameters to custom middleware, but I figured linking to the `createMiddleware` factory would give a user the answer they are looking for.

Let me know if you want any updates or changes!

closes: #317

### DIFF
--- a/guides/middleware.md
+++ b/guides/middleware.md
@@ -146,8 +146,16 @@ const logger = createMiddleware(async (c, next) => {
   console.log(`[${c.req.method}] ${c.req.url}`)
   await next()
 })
-
 ```
+:::info
+Type generics can be used with `createMiddleware`:
+
+```ts
+createMiddleware<{Bindings: Bindings}>(async (c, next) =>
+```
+:::
+
+### Modify the Response After Next
 Additionally, middleware can be designed to modify responses if necessary:
 
 ```ts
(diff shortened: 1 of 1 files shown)


guides/middleware.md
```diff
@@ -133,6 +133,39 @@ app.use('/message/*', async (c, next) => {
...
+```ts
+import { createMiddleware } from 'hono/factory'
+
+const logger = createMiddleware(async (c, next) => {
+  console.log(`[${c.req.method}] ${c.req.url}`)
+  await next()
+})
+
+```
+Additionally, middleware can be designed to modify responses if necessary:
```
yusukebe, line 151: I think it's better to separate this description from here and make it a sub-section of "Custom Middleware," which is put in.  The title will be like "Modify the Response after the `next()`".

NicoPlyley: Sounds good, I will update that soon

NicoPlyley: Fixed: 3abe563

yusukebe: Cool!

yusukebe (approved):
LGTM!

yusukebe:
@NicoPlyley 

Thanks! I'll merge now.

PR #323, opened 2024-04-30
feat: kawaii mode for mobile

neverclear86:
Hi @yusukebe 

I was sad that I couldn't see the kawaii logo when I activated kawaii mode on mobile, so I made it visible.

image


yusukebe:
Awesome! Thanks.

Issue #135, opened 2023-10-27
add `yarn`, `pnpm`, and `bun` for a package manager


yusukebe:
Currently, it using only `npm` like:

<img width="413" alt="Screenshot 2023-10-27 at 15 58 38" src="https://github.com/honojs/website/assets/10682/8ca1523e-c520-4d3b-81cc-d8e5b995cea8">

It should support `yarn`, `pnpm`, and `bun`.

PR #324, opened 2024-05-01
docs: remove mimes section in Node.js

ryuapp:
This issue was asked on `quick-questions` on Discord.
`mimes` option has not been added to `@hono/node-server` yet, so we will temporarily remove it from the documentation.

Ref: https://github.com/honojs/node-server/pull/135

### DIFF
--- a/getting-started/nodejs.md
+++ b/getting-started/nodejs.md
@@ -160,22 +160,6 @@ app.get(
 )
 ```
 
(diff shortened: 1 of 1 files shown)


yusukebe:
@ryuapp Thanks!

For #135, some users are still using Hono v3 now; we have to wait for a little for merging.

PR #339, opened 2024-05-05
docs(rpc): Add "be" because it is missing.

yasuaki640:
I added "be" because "able" should be treated as an adjective.


yusukebe:
Thanks!

PR #351, opened 2024-05-14
add warning about Bun's default request body limit

BobbyRaduloff:
In reference to honojs/hono#2671,

Bun terminates requests above 128MiB by default. Setting the Hono `bodyLimit` middleware to a higher value will not help as the request is terminated before it even reaches Hono. Additionally, Bun does not provide an explicit error about this and the `onError` handler on Hono's `bodyLimit` middleware doesn't get called either, making errors hard to debug.

A section is added to warn users about this behavior with mitigation steps.


yusukebe:
Hey @BobbyRaduloff 

Good!! Perfect. I'll merge this now.

PR #354, opened 2024-05-19
Dev Helper with Options Object

mountainash:
This change shows how the options object can be added (without it, this page doesn't show how to add the options)


yusukebe:
Hi @mountainash 

> without it, this page doesn't show how to add the options

I see! This is a good improvement. I'll merge it now.

PR #367, opened 2024-05-27
fix: fix other code block can't display

PaiJi:
image
 
the `code-group` cause the rest of code blocks in the page can't work.


yusukebe:
@PaiJi 

Thanks!

PR #371, opened 2024-05-28
Remove IP Section in Bun

NicoPlyley:
This becomes unneeded with the new 4.4.0 update. I was considering adding the connInfo instead, but considering other pages, it would appear to just be duplicated content


yusukebe:
Hi @NicoPlyley 

Yes, it can be removed. Thanks!

PR #375, opened 2024-05-30
feat: Replacement to TSDoc for API documentation

6km:
As [proposed by @yusukebe (Replacement to TSDoc for API documentation)(https://github.com/orgs/honojs/discussions/2854).

I will be working on implementing this in the meantime.

<br />

Main goals:
- API references should have a similar style and format to the Docs.
- For faster maintenance, auto-generate with every new build of the site.

<br />

API docs automatic generation process:
- Clone source code from honojs/hono in a temporary folder
- Run `typedoc` to generate the docs
- Delete source code files of Hono
- *Run the usual build process...*

<br />

Thanks for discussing the proposal:
- @goisaki 
- @yusukebe 
- @NicoPlyley

### DIFF
--- a/.vitepress/config.ts
+++ b/.vitepress/config.ts
@@ -232,6 +232,7 @@ export default defineConfig({
         'Copyright © 2022-present Yusuke Wada & Hono contributors. "kawaii" logo is created by SAWARATSUKI.',
     },
     nav: [
+      { text: 'API', link: '/api' },
       { text: 'Docs', link: '/top' },
       { text: 'Snippets', link: '/snippets/top' },
       { text: 'Examples', link: 'https://github.com/honojs/examples' },
--- a/.vitepress/config.ts
+++ b/.vitepress/config.ts
@@ -1,5 +1,6 @@
-import { defineConfig } from 'vitepress'
 import type { DefaultTheme } from 'vitepress'
+import { defineConfig } from 'vitepress'
+import typedocSidebar from '../docs/api/typedoc-sidebar.json'
 
 const sidebars = (): DefaultTheme.SidebarItem[] => [
   {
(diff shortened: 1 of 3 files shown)


6km:
Hi @yusukebe

What do you think about the API docs generation process?

yusukebe:
@6km 

Yeah. I also think in a similar way.

> Clone source code from honojs/hono in a temporary folder

I think we cah use like Git submodule feature.

6km:
@yusukebe Can we use tiged to clone the soruce code from honojs/hono? It's used by create-hono to clone started templates too!

goisaki:
Hi @yusukebe and @6km 
I don't understand the need to use a submodule, so I think this is a simple way to do it:
```yaml
- name: Checkout
  uses: actions/checkout@v4
  with:
    path: website

- name: Checkout code repo
  uses: actions/checkout@v4
  with:
    repository: honojs/hono
    path: hono
```
In this case, the next challenge is
- How do we run workflow as we upgrade?
  - Manual (not good)
  - Automatic (good, but hard)

yusukebe:
@goisaki 

> I don't understand the need to use a submodule, so I think this is a simple way to do it:

Yeah, you are right. I thought it is necessay for local development to refer the `honojs/hono` repository, but we can specify the path manually or an environment variable.

> In this case, the next challenge is

Yes. Our challenge is to create the GitHub actions to automate updating the docs.

yusukebe:
The ideal flow is the following:

1. `honojs/hono` main branch will be updated.
2. This `honojs/website` repo will receive the notification.
3. Check out the `honojs/hono` repo in GitHub Actions.
4. Build the docs in GitHub Actions.
5. `git add`, `git commit`, and `git push` the changes.
6. The website `https://hono.dev` will be updated.

It's difficult to try it in the current repo, so I think we have to do them in each private repo.

PR #378, opened 2024-06-01
feat: add Bun install on existing project

PlayerNguyen:
Add Bun install on an existing project with `bun install` command as an additional step.

### DIFF
--- a/getting-started/bun.md
+++ b/getting-started/bun.md
@@ -9,6 +9,7 @@ To install `bun` command, follow the instruction in [the official web site](http
 
 ## 2. Setup
 
+### 2.1. Setup a new project
 A starter for Bun is available. Start your project with "bun create" command.
 Select `bun` template for this example.
 
@@ -21,6 +22,13 @@ Move into my-app and install the dependencies.
 ```sh
 cd my-app
 bun install
+```
+
+### 2.2. Setup an existing project
+On an existing Bun project, we only need to install `hono` dependencies on the project root directory via
+
+```sh
+bun install hono
 ```
 
 ## 3. Hello World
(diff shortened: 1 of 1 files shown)


ryuapp:
Hello.
The purpose of getting started is to create a new Hono project, so this addition seems unnecessary.
The guide should be simple.

However, your point is correct, so why don't u add it as an other section(add Hono to an exsisting project)?

PlayerNguyen:
Hi @ryuapp, 

I understand that we can put this into a different section. However, I am not too familiar with this documentation format. Moreover, this one, I expected worked and tested with `Bun`. So the only question here is "It is necessary to create a new section instead?".

PlayerNguyen:
I found that the [Getting Started > Basic ](https://hono.dev/getting-started/basic) is quite suitable for putting the section above.

getting-started/bun.md
```diff
@@ -23,6 +24,13 @@ cd my-app
 bun install
 ```
 
+### 2.2. Setup an existing project
+On an existing Bun project, we only need to install `hono` dependencies on the project root directory via
+
+```sh
+bun install hono
```
yusukebe, line 31: I think [`bun add` ](https://bun.sh/guides/install/add) is better. What do you think of it?

yusukebe:
Hi @PlayerNguyen !

This is a good suggestion. I've left a comment. Check it.

getting-started/bun.md
PlayerNguyen: Sure, I am going to change it.

PlayerNguyen:
@yusukebe I already use `bun add` instead. Can you please check it again?

yusukebe (approved):
LGTM!

yusukebe:
@PlayerNguyen Looks good! Merging now. Thank you for your contribution!

PR #380, opened 2024-06-01
docs: improve cloudflare-workers page

NaingAungLuu:
Highlighted the information on module worker vs service worker regarding the local variables. The information wasn't highlighted so that new users of cloudflare worker runtime can find why their local environment variables aren't working.


yusukebe:
It makes sense for me. Thank you!

PR #383, opened 2024-06-06
docs: reword cloudflare-workers section on service worker mode vs module worker mode

WillSmithTE:
put module workers first since they're the recommended way 
move justification into the same paragraph to make sure it gets read

I had the same problem as [this commenter](https://github.com/honojs/hono/issues/194#issuecomment-2143549646). To me, this is clearer and points users more clearly to the happy path


yusukebe:
@WillSmithTE 

You are right. This should be first. Thanks!

PR #387, opened 2024-06-09
fix typo in cache.md

ykyki:
This PR fixes minor typos in `cache.md`.
https://hono.dev/middleware/builtin/cache

### DIFF
--- a/middleware/builtin/cache.md
+++ b/middleware/builtin/cache.md
@@ -20,7 +20,7 @@ import { cache } from 'hono/cache'
 ::: code-group
 
 ```ts [Cloudflare Workers]
-app.get(
+app.use(
   '*',
   cache({
     cacheName: 'my-app',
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @ykyki 

This is not a typo. It works well with `app.get()`.

ykyki:
Thanks @yusukebe .

After a quick check, I found that `app.get()` worked well.

However, `app.use()` also seems to work correctly.
I was under the impression that `app.use()` was more suitable for middlewares.

Is there any criterion for determining  which approach is more appropriate?

yusukebe:
Hi @ykyki 

I intended to apply cache middleware only for GET requests in this document, which is useful for real-world examples.

ykyki:
@yusukebe 

I understood your intention.
In light of that, `app.get()` is indeed more appropriate than `app.use()` for this case.

I intend to withdraw this PR.
Thank you for your guidance.

PR #390, opened 2024-06-15
docs: use JSR instead of deno.land/x

ryuapp:
Hono is no longer updated on deno.land/x, so change to JSR.
Since Deno also recommends JSR, use JSR instead of deno.land.

### DIFF
--- a/guides/middleware.md
+++ b/guides/middleware.md
@@ -102,8 +102,8 @@ app.use(
 In Deno, it is possible to use a different version of middleware than the Hono version, but this can lead to bugs.
 For example, this code is not working because the version is different.
 ```ts
-import { Hono } from 'https://deno.land/x/hono@v2.0.0/mod.ts'
-import { upgradeWebSocket } from 'https://deno.land/x/hono@v4.1.4/helper.ts'
+import { Hono } from 'jsr:@hono/hono@4.4.0'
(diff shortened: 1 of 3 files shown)


guides/middleware.md
```diff
@@ -102,8 +102,8 @@ app.use(
 In Deno, it is possible to use a different version of middleware than the Hono version, but this can lead to bugs.
 For example, this code is not working because the version is different.
 ```ts
-import { Hono } from 'https://deno.land/x/hono@v2.0.0/mod.ts'
-import { upgradeWebSocket } from 'https://deno.land/x/hono@v4.1.4/helper.ts'
+import { Hono } from 'jsr:@hono/hono@4.4.0'
```
yusukebe, line 105: It would be better to unify the `hono`'s versions with the following. So it should be `4.4.5`.

yusukebe:
@ryuapp 

Thanks! I've left the comments.

guides/middleware.md
ryuapp: This document introduces a bug that occur due to different versions.
So the different versions are correct.

yusukebe: Ahhh, sorry!

yusukebe:
@ryuapp Sorry for my wrong comment! Merging now.

PR #391, opened 2024-06-15
docs: use hono/bun instead in adapter example

ryuapp:
serveStatic() for Cloudflare-workers is deprecated, so change to Bun instead.
I thought of using websocket as an example, but I thought serveStatic() would be easier to understand, so I changed it to Bun.


ryuapp:
Since this chapter used Cloudflare workers as an example, I close this PR.

PR #392, opened 2024-06-15
docs: use `upgradeWebSocket()` instead in adapter example

ryuapp:
A deprecated function is used in the example, this PR fixes it.

Ref: #391


yusukebe:
@ryuapp 

Great! This is what I wanted to do!

PR #393, opened 2024-06-15
docs: improve text color in light mode

ryuapp:
The text color in light mode is pure black(`#000000`), which is not easy on the eyes.
Could you please make it the black color that is also used in [VitePress](https://vitepress.dev/)?

before:
image
after:
image

### DIFF
--- a/.vitepress/theme/custom.css
+++ b/.vitepress/theme/custom.css
@@ -4,7 +4,7 @@
 
   --vp-c-gray: #8e8e93;
 
-  --vp-c-text-light-1: #3c3c43;
+  --vp-c-text-light-1: #1f2328;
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @ryuapp 

Changing the color is a good idea, but I prefer a slightly darker color. How about using `#1f2328` instead of `#3c3c43`? The color is used on the GitHub websites.

ryuapp:
@yusukebe 

> Changing the color is a good idea, but I prefer a slightly darker color. How about using #1f2328 instead of #3c3c43? The color is used on the GitHub websites.

I agree with you since my aim is to avoid pure black. It has been fixed.

yusukebe:
@ryuapp 

Thanks! Let's go with this.

PR #406, opened 2024-06-19
Phrasing correction in benchmarks.md

davidwrossiter:
Minor adjustment of phrasing to make it the same as the example above.


yusukebe:
@davidwrossiter 

Thanks!

PR #411, opened 2024-06-22
ci: setup ci workflow

Jayllyz:
Created a CI workflow to check Prettier format and run build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,30 @@
+name: ci
+on:
+  push:
+    branches: 
+      - main
+  pull_request:
+
+env:
+  CI: true
+
+jobs:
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @Jayllyz 

Checking formatting is very good. But honestly, I think this CI is not necessary in my opinion. If some user fix the typo and create a PR which is not formatted. Then, the CI will fail, but if the user does not notice it, I will have to ask the user to fix the format to pass the CI. It's a burden for me. So, it will be good I or some bot fix it not related PRs.

Jayllyz:
No problem, I understand, thank you for your time and review.

yusukebe:
@Jayllyz 

Thank you for understanding! I'll close this PR now.

PR #416, opened 2024-06-25
Testing: Add Context section

JoaquimLey:
Also adds an example with custom routing.

### DIFF
--- a/docs/guides/testing.md
+++ b/docs/guides/testing.md
@@ -128,3 +128,59 @@ test('GET /posts', async () => {
   const res = await app.request('/posts', {}, MOCK_ENV)
 })
 ```
+
+## Context
+
+To set `c.set()` for testing, you can create a new app and add a custom testing middleware.
+
+For example, let's add a mock user object to "jwt" to all routes:
+
+```ts
+// Create a mock object
+const mockUser = {
+  id: "550e8400-e29b-41d4-a716-446655440000",
+  email: "johndoe@example.com",
+  age: 26
+};
+
+let app: Hono;
+
+beforeEach(() => {
+  app = new Hono();
+  
+  // Set any test data to your context
+  app.use("*", async (c, next) => {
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @JoaquimLey thank you for the PR. I'll comment here.

> To set values into your context `c.set()` for testing, create a custom testing middleware and add it before your test(s) run.

Regarding this, there is not only a case in which we have to set it **before** each test. If it does not have to initialize something, we can write it without `beforeEach`.

```ts
const app = new Hono<{
  Variables: {
    user: typeof mockUser
  }
}>()

app.use('*', async (c, next) => {
  c.set('user', mockUser)
  await next()
})

const getFriend = vi.fn()

app.get('/user', (c) => {
  const user = c.get('user')
  getFriend(user.id)
  return c.text('/user')
})

describe('Basic', () => {
  it('should return 200 response', async () => {
    const res = await app.request('/user')
    expect(res.status).toBe(200)
    expect(getFriend).toHaveBeenCalledWith(mockUser.id)
  })
})
```

The problem we have to say is the below case. This test will throw the error `Error: Can not add a route since the matcher is already built.`.

```ts
import { Hono } from 'hono'

const adminApp = new Hono()
adminApp.get('/', (c) => c.text('Admin Top'))

const app = new Hono()

beforeEach(() => {
  app.route('/admin', adminApp)
})

describe('Admin route', () => {
  it('should return 200 response', async () => {
    const res = await app.request('/admin')
    expect(res.status).toBe(200)
  })
  it('should return 404 response', async () => {
    const res = await app.request('/admin/foo')
    expect(res.status).toBe(404)
  })
})
```

To prevent that we have to write the `beforeEach`:

```ts
let app: Hono

beforeEach(() => {
  app = new Hono()
  app.route('/admin', adminApp)
})
```

I think what we have to write the document is this case.

PR #419, opened 2024-06-26
fix context naming in lambda example

fuxi-de:
The Lambda example on the website uses a different property name than the ones actually provided at runtime (`context` vs. `lambdaContext`). 

This PR fixes the naming in the example, so it is obvious for new users.


yusukebe (approved):
LGTM!

yusukebe:
@fuxi-de Thanks!

PR #421, opened 2024-06-26
fix typo in Deno docs

brettchalupa:


yusukebe:
@brettchalupa 

Thanks!

PR #422, opened 2024-06-27
hc does not support file uploading

yusukebe:


movahhedi:
I am literally using `hc` to upload files! how is this not possible? With `"form"`.


#### client
```ts
const submitResponse = await client.api.entity["article-image"][":id"].edit.$post(
	{
		param: { id: "" + thisId },
		form: {
			file: new File([fileToUpload], filename, {
				type: fileToUpload.type,
			}),
		},
	},
);
```

#### server
```ts
.post(
	"/:id/edit",
	zValidator(
		"form",
		z.object({
			file: z.instanceof(File).optional(),
		}),
	),
	// ...
```

yusukebe:
@movahhedi 

I think you can do it! Isn't it working well?

movahhedi:
Yes, and brilliantly. I don't know why you added this note. I think it should say "File uploading is currently only available using `form`, and provide the example i sent below it.

yusukebe:
@movahhedi 

> I don't know why you added this note.

Honestly, I also don't remove why I add this note:)

So, removing this note and adding an example like the one you wrote would be good.

PR #426, opened 2024-07-03
remove Deno Docs from use-cases

ryuapp:
Deno Docs has been moved to Lume, so Hono is no longer used.
https://github.com/denoland/deno-docs/pull/519


yusukebe:
@ryuapp 

Ah, Lume! I see! Thanks for sharing it.

PR #429, opened 2024-07-04
docs: improve options section

### LINKED ISSUE #403: Updating Options
Hi @yusukebe,

I wanted to discuss updating how options are displayed on the site. Currently, they are just a list that looks a little crowded.

I am proposing to update it to be table-based.

Screenshot 2024-06-18 at 3 32 45 PM

I think the table option is easier to understand and makes the docs look cleaner. However, on mobile, this means the table would be scrollable and not displayed in full like it is now. 

Screenshot 2024-06-18 at 3 33 13 PM

I don't think it is a huge issue since it's still readable with a scroll, which is an overflow scroll, it does not affect the horizontal width of the window. And I would think the majority of users are using a desktop device to read the documentation.

Based on what you think or if you want to try it yourself, I will submit a PR. Lastly, should we open a discussion for the website and ideas? All previous conversations are in an issue on the Hono repo

ryuapp:
closes #403

I made a few changes from the design proposed in the issue.
I set types as a code block and placed either required or optional at the beginning.
Basically, types are expressed as blocks of code, so I think that's natural.

Light
image

Dark
image

If you have any opinions about the design, please feel free to comment.

### DIFF
--- a/.vitepress/theme/custom.css
+++ b/.vitepress/theme/custom.css
@@ -18,6 +18,11 @@
   --vp-c-orange-lighter: #ff9c24;
   --vp-c-orange-dark: #e07700;
   --vp-c-orange-darker: #d68e22;
+
+  --vp-c-badge-green-light: #308b58;
+  --vp-c-badge-green-light-bg: #bff8d7;
+  --vp-c-badge-green-dark: #1fd85d;
+  --vp-c-badge-green-dark-bg: #308b5769;
 }
 
 /**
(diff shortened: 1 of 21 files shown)


ryuapp:
- The cookie/ssg helper options is currently easier to see than this proposal, so I have put it on hold.
- Unify the notation of default values. `The default is ...`

ryuapp:
Fix #449

Hello @yusukebe .
This PR is available for review. Can you check this when you're free?

yusukebe:
Okay! I’ll do it later.

yusukebe:
Hi @ryuapp 

I've checked. This is awesome!! I love it. There is no problem, and it looks good to me!

@NicoPlyley @goisaki Can you see this?

goisaki:
> Hi @ryuapp 
> 
> 
> 
> I've checked. This is awesome!! I love it. There is no problem, and it looks good to me!
> 
> 
> 
> @NicoPlyley @goisaki Can you see this? 
I think so too. Awesome!

yusukebe:
@ryuapp 

It's ready to merge, and do it now. Thanks!

PR #431, opened 2024-07-06
Add Deno Deploy compresses the body response

gmarty:
Add Deno Deploy to the list of services that already compress the body response.

See https://docs.deno.com/deploy/api/compression/


yusukebe:
Hi @gmarty 

Good! Thanks.

PR #435, opened 2024-07-13
fix pylon example

kazu728:
Added async await to `createPost` because the following error occured.

> Property '$createPost' does not exist on type `Promise<User>`.


yusukebe:
@kazu728 

Thanks!

PR #436, opened 2024-07-13
docs (rpc): multiple parameters

Jay-Karia:
Add a section under `Path Parameters`, to show how to call API with endpoints like: `/posts/:postId/:authorId` using hono client.

### DIFF
--- a/docs/guides/rpc.md
+++ b/docs/guides/rpc.md
@@ -176,6 +176,41 @@ const res = await client.posts[':id'].$get({
 })
 ```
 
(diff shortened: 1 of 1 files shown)


eofreternal:
I think this would be a good addition to the docs. I had to figure this out for myself, so this would help future devs

yusukebe (approved):

yusukebe:
@Jay-Karia @eofreternal Thanks!

I updated the branch myself, and it looks good.

PR #451, opened 2024-07-21
docs(cors): add documentation for environment-dependent CORS configuration

3w36zj6:
close https://github.com/honojs/hono/issues/1133

IMO: Given that this insight can be leveraged with other middleware as well, it might be a good idea to include it on the [Middleware page](https://hono.dev/docs/guides/middleware) as well.

### DIFF
--- a/docs/middleware/builtin/cors.md
+++ b/docs/middleware/builtin/cors.md
@@ -85,3 +85,16 @@ The value of "_Access-Control-Allow-Credentials_" CORS header.
 ### <Badge type="info" text="optional" /> exposeHeaders: `string[]`
 
 The value of "_Access-Control-Expose-Headers_" CORS header. The default is `[]`.
+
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @3w36zj6 

Thank you for the PR.

> IMO: Given that this insight can be leveraged with other middleware as well, it might be a good idea to include it on the [Middleware page](https://hono.dev/docs/guides/middleware) as well.

This will be great. What we have to do may be update the [Middleware page](https://hono.dev/docs/guides/middleware) instead of the [CORS page](https://hono.dev/docs/middleware/builtin/cors). If you want to, can you create another PR to update the Middleware page?

3w36zj6:
I've created insights for general middleware in #452. Therefore, I've made more specific revisions to the example on the CORS page.

yusukebe:
@3w36zj6 Nice! Merging.

PR #454, opened 2024-07-24
docs: remove an expired link of serveStatic example for cloudflare-workers

ayanami77:
I found that an example for `serveStatic` for cloudflare-workers does not exist in the honojs/examples repo, so I deleted this line.

```md
See [Example](https://github.com/honojs/examples/tree/main/serve-static).
```


yusukebe:
@ayanami77 Thanks!

PR #460, opened 2024-08-23
docs(middleware): add instructions for extending context

iveshenry18:
Spent an hour trying to figure this out, then found the answer buried in [this discussion](https://github.com/honojs/hono/issues/414#issuecomment-1800878459). Looks like there's still ongoing discussion re: the ergonomics of this, but I figured it'd be useful to surface the current method.


yusukebe:
Hi @iveshenry18 

It's good to write that down. Thank you! Merging.

PR #462, opened 2024-08-28
fix: incorrect function notation in ssgParams section

taga3s:
## description

There is no function named `generateStaticPaths`, according to the Next.js docs. So, based on the `ssgParams` example of the Hono, I correct it to `generateStaticParams`). (ref: [Functions: generateStaticParams | Next.js](https://nextjs.org/docs/app/api-reference/functions/generate-static-params))


yusukebe (approved):
LGTM!

yusukebe:
@taga3s Thanks!

PR #467, opened 2024-09-01
docs(cache): can be used with Deno Deploy

ryuapp:
Add that cache middleware is can be used in Deno Deploy.
Ref: https://deno.com/blog/deploy-cache-api

In addition, the `Cache-Control` header is only respected by Cloudflare Workers, so state that.


yusukebe:
@ryuapp 

I thought we needed to do it. Thank you!

PR #487, opened 2024-09-12
Update Cloudflare Testing and Cloudflare Workers Getting started

JagritGumber:
- Cloudflare Testing example has been added in the examples section
- Cloudflare Workers getting starter has been updated to include the updated testing library instead of the outdated one
- The documentation examples have been updated


yusukebe:
Hi @JagritGumber 

Sorry for the late review. This is great! It will be helpful. But one thing. Is it necessary to write the D1 things, including "migrations"? I think it would be better to make this section simple and add a link to Cloudflare's official docs with descriptions. What do you think of it?

JagritGumber:
That actually makes sense I may have overlooked the possibility that the users might integrate different databases like neon, supabase etc, surely I'd work on that, try to remove the d1 specific details and rather link to cloudflare so they can get in depth knowledge over there.

JagritGumber:
Hi @yusukebe 
I made the necessary changes to make the example more versatile and less focused on D1 specific details that makes it overall better as an example.

yusukebe:
@JagritGumber 

Thank you! It looks good. But I have one thing I would like you to add. How about adding the case of using `app.request()`? It's okay to use `hono/testing`, which has great Type inferring, but using `app.request()` is an easy one.

yusukebe:
Hi @JagritGumber 

Thank you! I've updated the doc to make it easy to understand and correct some stuff. Let's go with it. I appreciate your help!

PR #496, opened 2024-09-17
Add warning about validating headers in validation.md

sushichan044:
added warning about header name case.
ref: https://github.com/honojs/hono/issues/3416#issuecomment-2354669238

I'm considering adding a few more reasons for this behavior.


yusukebe (approved):
LGTM!

yusukebe:
@sushichan044 

Thanks! This is a good description which includes a practical example. I'll merge this now.

PR #502, opened 2024-09-30
Fix setRenderer examples

freetonik:


yusukebe:
Hi @freetonik 

This is a JSX file. It does not need quotes.

PR #508, opened 2024-10-05
Update websocket.md - fix Bun examples

PizzaConsole:


yusukebe:
Hi @PizzaConsole 

`Bun.serve` is not wrong in this case, but it may be good to unify `export default` for the whole website. It uses `export default` in other places. I'll merge this. Thanks!

PR #510, opened 2024-10-07
Add reverse proxy example

iflamed:


yusukebe:
Hi @iflamed 

Try this:

```ts
app.all('*', (c) => {
  const url = new URL(c.req.path, 'https://www.example.com')
  return fetch(url, c.req.raw)
})
```

yusukebe:
@iflamed 

I haven't tried it yet, but I think here:

```ts
return await fetch(upstream, {
  method: c.req.raw.method,
  body: c.req.raw.body,
  credentials: c.req.raw.credentials,
  cache: c.req.raw.cache,
  headers: c.req.raw.headers,
  referrer: c.req.raw.referrer,
  referrerPolicy: c.req.raw.referrerPolicy,
  integrity: c.req.raw.integrity,
  keepalive: false,
  mode: c.req.raw.mode,
  redirect: 'manual'
})
```

can be written like this:

```ts
return await fetch(upstream, {
  ...c.req.raw,
  keepalive: false,
  redirect: 'manual'
})
```

iflamed:
@yusukebe yes, some attributes not supported by Cloudflare pages.

yusukebe:
@iflamed 

Ah. We have to remove some properties from `c.req.raw`, right?

iflamed:
@yusukebe yes, and I found my code not support post request, will report content-length not match.

PR #511, opened 2024-10-07
Update routing.md

ecwyne:
The example given was incorrect.


yusukebe:
@ecwyne 

You are right. It was incorrect. Thanks!

PR #512, opened 2024-10-09
Update stacks.md

subhamnayak76:
need add "export default app" to run else showing error.


yusukebe:
Hi @subhamnayak76 

Thank you for the PR. It's intended that it does not write `export default app` because some platforms do not use it. For example, some are using `export const handler = handle(app)`.

PR #517, opened 2024-10-21
Update best-practices.md

enyelsequeira:
I tested this out, when building in a monorepo setup and noticed that when you don't chain one method after another you lose the typesafety you get when using `hc<appType>`

### DIFF
--- a/docs/guides/best-practices.md
+++ b/docs/guides/best-practices.md
@@ -110,3 +110,21 @@ app.route('/books', books)
 
 export default app
 ```
+The above code would work fine, however doing this would lose the typesafe
+
+a better solution would be something like this
+
+```ts
+// authors.ts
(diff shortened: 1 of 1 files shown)


docs/guides/best-practices.md
```diff
@@ -110,3 +110,21 @@ app.route('/books', books)
 
 export default app
 ```
+The above code would work fine, however doing this would lose the typesafe
```
yusukebe, line 113: Can you add the following sentence or a similar one?

> "If you want to use [the RPC feature](https://hono.dev/docs/guides/rpc)"

enyelsequeira: of course 
if you want to use `RPC` features a better solution would be chaining the methods as described below

PR #521, opened 2024-10-26
docs(api): add missing `newResponse()` context

garysassano:


yusukebe:
Hi @garysassano 

In this case, we intend to use the [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object directly. So this change is not necessary.

garysassano:
Why use the `Response` object? Wasn't the `c.newResponse` context created as an alternative to uniform the syntax?

PR #524, opened 2024-11-04
docs(serve-static/node): fix middleware invocation

geoffturk:
Fixing the middleware invocation example for the `serveStatic` docs for Node...

### DIFF
--- a/docs/getting-started/nodejs.md
+++ b/docs/getting-started/nodejs.md
@@ -145,7 +145,7 @@ You can use `serveStatic` to serve static files from the local file system.
 ```ts
 import { serveStatic } from '@hono/node-server/serve-static'
 
-app.use('/static/*', serveStatic({ root: './' }))
+app.use('*', serveStatic({ root: './static' }))
 ```
 
 ### `rewriteRequestPath`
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @geoffturk 

The current example shows that specifying the serve static path as `/static/*` is good. This PR is not necessary.

geoffturk:
@yusukebe My apologies for not being clearer in my original message.

My intention is to be able to put static files like `favicon.ico` or `images/logo.png` in the `static` directory (in the root of the project) and have them served up when the page is loaded in a browser.

`app.use('/static/*', serveStatic({ root: './' }))` does not achieve that but `app.use('*', serveStatic({ root: './static' }))` does.

So I think you need to update the code example in the docs. Please let me know if this is still not clear. Thanks.

yusukebe:
@geoffturk 

Thank you for the explanation. Then, we have to add the description for the some patterns like this Deno-case:

https://hono.dev/docs/getting-started/deno#serve-static-files

geoffturk:
@yusukebe OK, please see this latest commit.

yusukebe:
@geoffturk Thanks.

As you know, some concepts are mixed in that description. How about like this #525 ?

geoffturk:
@yusukebe Yes, #525 looks really good. More concise and covers both scenarios. It might be worth plugging the same content into the Bun and Deno "Getting Started" pages as well in case readers never look at the Node.js page. Thanks for your time looking into this. It definitely makes the docs clearer! 💯

PR #528, opened 2024-11-10
docs(getting-started/vercel): add Node.js runtime usage for App Router

### LINKED ISSUE #529: docs: Missing Node.js runtime documentation for App Router
The current Vercel documentation only shows how to use Node.js runtime with Page Router, but misses the simpler implementation for App Router users.

### Current situation:
- Only Page Router implementation is documented
- Shows installation of @hono/node-server which isn't needed for App Router

### Needed changes:
- Add documentation for App Router Node.js runtime usage
- Show that only `export const runtime = 'nodejs'` is needed
- Organize the section to clearly separate both approaches

farhadham:
This PR adds documentation for using Hono with Next.js Node.js runtime in App Router setup.

Changes:
- Add new section showing how to use Node.js runtime with App Router
- Reorganize the Node.js section to clearly separate App Router and Page Router implementations
- Document that App Router doesn't require @hono/node-server package
- Maintain existing Page Router implementation details

The App Router implementation is simpler as it only requires setting `runtime = 'nodejs'` without additional packages.

Closes #529

### DIFF
--- a/docs/getting-started/vercel.md
+++ b/docs/getting-started/vercel.md
@@ -141,13 +141,37 @@ If you have a Vercel account, you can deploy by linking the Git repository.
 
 You can also run Hono on Next.js running on the Node.js runtime.
 
-First, install the Node.js adapter.
+### App Router
+
+For the App Router, you can simply set the runtime to `nodejs` in your route handler:
+
+```ts
+import { Hono } from 'hono'
+import { handle } from 'hono/vercel'
+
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @farhadham 

Awesome! It will be clear. I've updated the docs for formatting. Let's go with this.

farhadham:
Hi @yusukebe !
It was my Honor doing anything for this project

PR #546, opened 2024-12-08
docs: add Google App Engine getting started guide

ysknsid25:
I like Google Cloud. Of course, I also like Hono.

But I can't find anything related to Google Cloud on the Getting Start page, even though there are AWS, Lambda, and Azure Functions...

I am very sad because I like Google Cloud.

So I described what I need to do to get Hono running on App Engine in [the advent calendar](https://zenn.dev/yskn_sid25/articles/1d6678105fa1ad). In addition, I thought I would add the Setup for running Hono on App Engine to the official Hono page.

I was wondering if this procedure could also be automated from the command? I am thinking that it might be possible to fix `starter` and `create-hono`.

We apologize for the inconvenience, and thank you for your understanding.

### DIFF
--- a/.vitepress/config.ts
+++ b/.vitepress/config.ts
@@ -70,6 +70,10 @@ const sidebars = (): DefaultTheme.SidebarItem[] => [
         link: '/docs/getting-started/service-worker',
       },
       { text: 'Node.js', link: '/docs/getting-started/nodejs' },
+      {
+        text: 'Google App Engine',
+        link: '/docs/getting-started/google-app-engine',
+      },
     ],
   },
   {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @ysknsid25 

The reason we didn't add Google App Engine to the list is that anyone who wants to use it can just look at Node.js. AWS Lambda and Lambda@Edge do not use Node.js, and they are specific platforms, so we need to put them on our website. For GAE, I don't think we need this PR because users just do it the Node.js way.

ysknsid25:
@yusukebe 

I see, I understand. I think you are right.
Close this PR. Thank you for your time and confirmation!

yusukebe:
@ysknsid25 

Thank you for spending time and understanding. I appreciate you using Hono!

PR #547, opened 2024-12-08
Fix linter warning (Deno)

bleetube:
Deno linter problem: 'c' is declared but its value is never read.

We fix this by not destructuring the context parameter. We could have instead prefixed it with an underscore, but since the purpose of this document is to remain simple and easy to consume, we preferred this approach. There might be some good reasons why this document would want to retain the context param, so feel free to close this if that's the case!


yusukebe (approved):
LGTM!

yusukebe:
Hi @bleetube 

Thank you for the PR. This is good because it's simple. I'll merge now!

PR #548, opened 2024-12-16
docs: add more details to cloudflare workers bindings section

jw-12138:


docs/getting-started/cloudflare-workers.md
```diff
@@ -247,6 +247,50 @@ app.put('/upload/:key', async (c, next) => {
...
+```ts
+import { Context } from 'hono'
+
+type Bindings = {
+  MY_BUCKET: R2Bucket
+  USERNAME: string
+  PASSWORD: string
+}
+
+export const handler = async (c: Context<{
```
yusukebe, line 261: We don't recommend using `Context` for `c` for the following reasons:

https://hono.dev/docs/guides/best-practices#don-t-make-controllers-when-possible

docs/getting-started/cloudflare-workers.md
```diff
@@ -247,6 +247,50 @@ app.put('/upload/:key', async (c, next) => {
...
+```sh
+wrangler types
+```
+
+After running the command you will get a `worker-configuration.d.ts` file which will have a typescript interface called `Env`.
+
+```ts
+// Generated by Wrangler by running `wrangler types`
+
+interface Env {
```
yusukebe, line 279: The naming `Env` will be confusing because we often use the following code:

```ts
type Env = {
  Bindings: {
    KV: KVNamespace
  },
  Variables: {
    foo: string
  }
}
```

Actually, the code which is generated from [C3](https://developers.cloudflare.com/pages/get-started/c3/) does not use `Env` but uses `CloudflareBindings`:

```ts
// Generated by Wrangler by running `wrangler types --env-interface CloudflareBindings`

interface CloudflareBindings {
	MY_VARIABLE: "production_value";
}
```

PR #550, opened 2024-12-18
Remove boundary assertion from regex example

mootari:
From what I understand, param regexes are always anchored. As such the second example is misleading as it implies the opposite.


yusukebe (approved):
LGTM!

yusukebe:
@mootari 

You are right! Thanks.

PR #552, opened 2024-12-19
docs: (RPC) add info about matching hono versions client-server

catalinpit:
Add information about using Hono RPC in projects where the client and server are in a single project. I spent way too much time wrestling with this before finding a comment online about the Hono version.

### DIFF
--- a/docs/guides/rpc.md
+++ b/docs/guides/rpc.md
@@ -491,6 +491,16 @@ This is a type instantiation for a single route. While the user doesn't need to 
 
 However, we have some tips to mitigate this issue.
 
+#### Hono version mismatch
+
+If your backend is separate from the frontend and lives in a different directory, you need to ensure that the Hono versions match. If you use one Hono version on the backend and another on the frontend, you'll run into issues such as "_Type instantiation is excessively deep and possibly infinite_".
+
+![hono-version-mismatch](https://github.com/user-attachments/assets/e4393c80-29dd-408d-93ab-d55c11ccca05)
+
+#### TypeScript project references
+
+Like in the case of [Hono version mismatch](#hono-version-mismatch), you'll run into issues if your backend and frontend are separate. If you want to access code from the backend (`AppType`, for example) on the frontend, you need to use [project references](https://www.typescriptlang.org/docs/handbook/project-references.html). TypeScript's project references allow one TypeScript codebase to access and use code from another TypeScript codebase.  *(source: [Hono RPC And TypeScript Project References](https://catalins.tech/hono-rpc-in-monorepos/))*.
+
 #### Compile your code before using it (recommended)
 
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @catalinpit !

Thank you for the PR. This will be helpful.

I think we must add the description, not be in the `Compile your code before using it (recommended)` section. You can create a new section about it at the same level as `Compile your code before using it (recommended)`:

CleanShot 2024-12-21 at 16 48 25@2x

Because the problem is not only when the code is being compiled; other problems will happen without compiling. For example, suppose you have monorepo, which includes one server application and one frontend. In that case, if the versions are different and, you use the exported type from the server app in the frontend, the editor throws the error like this:

CleanShot 2024-12-21 at 16 54 14@2x

Repro: https://github.com/yusukebe/hono-version-mismatch

This problem occurs not only due to a version mismatch but also when the paths to the `hono` files referenced by each project are different. However, in many cases, ensuring that the versions are the same can solve the problem.

catalinpit:
Hey @yusukebe, thanks for the reply and suggestion! I actually ran into that problem as well - `Type instantiation is excessively deep and possibly infinite`.

I'll add a new section!

catalinpit:
Done. Change made. 🙌

docs/guides/rpc.md
```diff
@@ -523,8 +533,6 @@ const res = await client.posts.$post({
 
 If your project is a monorepo, this solution does fit well. Using a tool like [`turborepo`](https://turbo.build/repo/docs), you can easily separate the server project and the client project and get better integration managing dependencies between them. Here is [a working example](https://github.com/m-shaka/hono-rpc-perf-tips-example).
 
-If your client and server are in a single project, [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) of `tsc` is a good option.
```
yusukebe, line 526: Should we remove this line? I think it's OK to leave it.

yusukebe:
@catalinpit 

Sorry for the late response. I've added the comment!

docs/guides/rpc.md
catalinpit: I added it above in its section `#### TypeScript project references`. Should I leave it here as well?

catalinpit:
> @catalinpit
> 
> Sorry for the late response. I've added the comment!

No worries. Thanks for the replies!

docs/guides/rpc.md
yusukebe: Ahhhh, I missed it! Let's go with this.

yusukebe (approved):
LGTM!

yusukebe:
@catalinpit 

Looks good. Merging now. Thank you!

catalinpit:
> @catalinpit
> 
> Looks good. Merging now. Thank you!

Thanks a bunch!

PR #554, opened 2024-12-21
Fix deno add command

brettchalupa:
The jsr: prefix is needed, otherwise the previous command fails with Deno 2.

Screenshot showing the error and then fix:

image


yusukebe (approved):
LGTM!

yusukebe:
@brettchalupa 

Thanks! Looks good to me.

PR #565, opened 2025-01-19
port number settings for Cloudflare

sakird:
Added instructions for changing port numbers for application developed for Cloudflare. 

Following this documentation page:
[Wrangler Configuration](https://developers.cloudflare.com/workers/wrangler/configuration/#local-development-settings)

### DIFF
--- a/docs/getting-started/cloudflare-pages.md
+++ b/docs/getting-started/cloudflare-pages.md
@@ -403,3 +403,16 @@ app.get('/hello', (c) => {
 
 export const onRequest = handle(app)
 ```
+
+## Change port number
+
+You can specify the port number by updating the contents of `wrangler.toml` like this::
+
+```toml
+[dev]
+ip = "192.168.1.1"
+port = 8080
+local_protocol = "http"
+```
+
+For more information: [Wrangler Configuration](https://developers.cloudflare.com/workers/wrangler/configuration/#local-development-settings)
--- a/docs/getting-started/cloudflare-workers.md
+++ b/docs/getting-started/cloudflare-workers.md
@@ -126,6 +126,19 @@ bun run deploy
 
 That's all!
 
+## Change port number
+
+You can specify the port number by updating the contents of `wrangler.toml` like this::
+
+```toml
+[dev]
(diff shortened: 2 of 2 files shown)


yusukebe:
Hi @sakird 

This is one of the configurations of Wrangler, not Hono-side. So I think we don't have to write about the setting of the port; it's ok to add only a link to the doc: https://developers.cloudflare.com/workers/wrangler/configuration/

Adding the line in `wrangler.toml` is not the only way to change it. You can change the port with the `wrangler dev --port` command, or you can change it by editing `wrangler.json` (JSON support is a new feature). So, it's better to add the link instead of writing details in `hono.dev`.

sakird:
Hi @yusukebe, 

I have updated the contents of this pull request to include only the links for configuration files and CLI options.

docs/getting-started/cloudflare-workers.md
```diff
@@ -126,6 +126,14 @@ bun run deploy
 
 That's all!
 
+## Change port number
```
yusukebe, line 129: This should be placed in the following category since it's for development.

CleanShot 2025-01-20 at 15 22 39@2x

docs/getting-started/cloudflare-pages.md
```diff
@@ -403,3 +403,11 @@ app.get('/hello', (c) => {
 
 export const onRequest = handle(app)
 ```
+
+## Change port number
```
yusukebe, line 407: Our starter template for Cloudflare Pages does not use a `wrangler` for the `dev` command such as `npm run dev`. You should not write this here.

docs/getting-started/cloudflare-workers.md
sakird: I have moved it under 3. Run as a sub content.

docs/getting-started/cloudflare-pages.md
sakird: Done.

yusukebe (approved):
LGTM!

yusukebe:
@sakird 

Thanks!

PR #567, opened 2025-01-24
Update cors.md to explain that CORS needs to be called before the route

jamiecarter7:
Explain that CORS should be called before the route is called


yusukebe (approved):
LGTM!

yusukebe:
Hi @jamiecarter7 

Looks good! This will be helpful. Thank you!

PR #576, opened 2025-02-02
feat: replaced the logo with an svg version

### LINKED ISSUE #574: [Need SVG logos] Only PNG logo assets
Hey devs can you add SVG logo version's that will be much easy to use and its better to use single asset 
as you love speed and performance it give some boost for sure and you also don't have to maintain  different size for same thing as PNG just saves you **524.49** KB and scales lossless. And SVG's are flexible to edit and developer friendly. 

Image

 
Just small performance boost every KB counts 😄

MathurAditya724:
Closed #574 

- [x] Changed the `.png` logo file with `.svg`
- [x] Updated the config file to use the new logo


yusukebe (approved):
LGTM!

yusukebe:
@MathurAditya724 

So quick! Thank you so much!

PR #577, opened 2025-02-03
Patching types of c.env

shahil-yadav:
Problem: TS2345: Argument of type Context<BlankEnv, "/ env", BlankInput> is not assignable to parameter of types

Patch: Directly applying binding types to the hono constructor fixes the types issue, i found out this in the cloudfare section of loading env


yusukebe:
Hi @shahil-yadav 

Thank you for the PR. This is a bug. I'll fix it.

PR #604, opened 2025-02-28
docs(env): mention wrangler.jsonc as a supported format in env for Cloudflare Workers

rohitaryal:


yusukebe:
@rohitaryal Thanks!

PR #605, opened 2025-03-02
Update basic.md to use latest deno syntax

lmtr0:
Hi there, deno has updated their CLI and the old syntax doesn't work, so I updated it to the new one :)

### DIFF
--- a/docs/getting-started/basic.md
+++ b/docs/getting-started/basic.md
@@ -25,7 +25,7 @@ bun create hono@latest my-app
 ```
 
 ```sh [deno]
-deno init --npm hono@latest my-app
+deno run npm:create-hono@latest my-app
 ```
 
 :::


yusukebe:
Hi @lmtr0 

Thank you for the PR. The current syntax `deno init ...` works fine with the latest Deno (v2.2.2). But should we replace it?

lmtr0:
Funny, when I tested this, It wasn't working properly, and errored out explaining that the run syntax I show in my PR was the correct way of running these scripts.

Now, however, I tested it again, after @yusukebe said it was working fine, and It worked without errors......
Anyway, if you would rather not merge this, not a problem, I just opened because It gave me an error when I tried running it a few hours ago

yusukebe:
@lmtr0 

Thank you for the comment and testing. Okay, in this case, holding the current command is better. Closing this PR, but I appreciate creating the PR.

PR #606, opened 2025-03-04
docs: add some details to address common questions

ambergristle:
## TL;DR
Made a few additions based on questions I've seen come up frequently on Discord. More can probably be done, but I wanted to keep changes contained. Let me know if I should have created an issue or something instead.

### Stacks
Added some details about RPC inference and chaining. In my opinion, the examples are a little confusing because they use `const route` for the RPC typing (which isn't really explained). I actually missed that the first time I read through, and initially thought the examples were actually breaking.
- https://discord.com/channels/1011308539819597844/1346424853196050472/1346438748342190141

  ```typescript
  const app = new Hono()
  // directly chaining here, and using `typeof app`,  
  // seems more consistent w other Hono docs, and
  // how many devs architect apps
  const route = app.get(/** */)
  // this name is also a little misleading
  export AppType = typeof route
  ```
### Getting Started: NodeJS
Moved build-prep steps before dockerfile example, and added a bit about Vite. I haven't seen much about the Vite plugins in the docs, or even links to the repo (except for ssg).
- https://discord.com/channels/1011308539819597844/1012485912409690122/1345994752197267539

### Middleware
Added a little info about getting middleware types in handlers. This seems to be a common source of confusion or misunderstanding, especially when it comes to how chaining vs injecting middleware affects typing.

### RPC
Specified that RPC cannot currently infer types from middleware, and linked the issue.

### DIFF
--- a/docs/guides/rpc.md
+++ b/docs/guides/rpc.md
@@ -2,7 +2,7 @@
 
 The RPC feature allows sharing of the API specifications between the server and the client.
 
-You can export the types of input type specified by the Validator and the output type emitted by `json()` when called in a handler. And Hono Client will be able to import it. At this time, responses returned from middleware are [not inferred by the client.](https://github.com/honojs/hono/issues/2719)
+Using the `AppType` exported from your server code, the Hono Client can infer both the input type(s) specified by the Validator, and the output type(s) emitted by handlers returning `c.json()`. At this time, responses returned from middleware are [not inferred by the client.](https://github.com/honojs/hono/issues/2719)
 
 > [!NOTE]  
 > For the RPC types to work properly in a monorepo, in both the Client's and Server's tsconfig.json files, set `"strict": true` in `compilerOptions`. [Read more.](https://github.com/honojs/hono/issues/2270#issuecomment-2143745118)
--- a/docs/guides/rpc.md
+++ b/docs/guides/rpc.md
@@ -1,8 +1,13 @@
 # RPC
 
-The RPC feature allows sharing of the API specifications between the server and the client.
+The RPC feature allows sharing of the API specifications between the server and the client. 
 
-Using the `AppType` exported from your server code, the Hono Client can infer both the input type(s) specified by the Validator, and the output type(s) emitted by handlers returning `c.json()`. At this time, responses returned from middleware are [not inferred by the client.](https://github.com/honojs/hono/issues/2719)
+First, export the `typeof` your Hono app (commonly called `AppType`)—or just the routes you want available to the client—from your server code.
+
+By accepting `AppType` as a generic parameter, the Hono Client can infer both the input type(s) specified by the Validator, and the output type(s) emitted by handlers returning `c.json()`. 
+
+> [!NOTE]
+> At this time, responses returned from middleware are [not inferred by the client.](https://github.com/honojs/hono/issues/2719)
 
 > [!NOTE]  
 > For the RPC types to work properly in a monorepo, in both the Client's and Server's tsconfig.json files, set `"strict": true` in `compilerOptions`. [Read more.](https://github.com/honojs/hono/issues/2270#issuecomment-2143745118)
--- a/docs/guides/rpc.md
+++ b/docs/guides/rpc.md
@@ -2,7 +2,7 @@
 
 The RPC feature allows sharing of the API specifications between the server and the client.
 
-You can export the types of input type specified by the Validator and the output type emitted by `json()`. And Hono Client will be able to import it.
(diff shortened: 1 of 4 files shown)


docs/guides/rpc.md
```diff
@@ -2,7 +2,7 @@
 
 The RPC feature allows sharing of the API specifications between the server and the client.
 
-You can export the types of input type specified by the Validator and the output type emitted by `json()`. And Hono Client will be able to import it.
+You can export the types of input type specified by the Validator and the output type emitted by `json()` when called in a handler. And Hono Client will be able to import it. At this time, responses returned from middleware are [not inferred by the client.](https://github.com/honojs/hono/issues/2719)
```
brettimus, line 5: nit: missing particle, "**the** input type"

```suggestion
You can export the types of both the input type specified by the Validator and the output type emitted by `json()` when called in a handler. And Hono Client will be able to import it. At this time, responses returned from middleware are [not inferred by the client.](https://github.com/honojs/hono/issues/2719)
```

ambergristle: more than a nit, i think. i tried to minimize changes, but i ended up overloading that sentence grammatically. 

i've revised the whole paragraph to try and reduce the cognitive load

docs/guides/rpc.md
```diff
@@ -2,7 +2,7 @@
 
 The RPC feature allows sharing of the API specifications between the server and the client.
 
-You can export the types of input type specified by the Validator and the output type emitted by `json()`. And Hono Client will be able to import it.
+Using the `AppType` exported from your server code, the Hono Client can infer both the input type(s) specified by the Validator, and the output type(s) emitted by handlers returning `c.json()`. At this time, responses returned from middleware are [not inferred by the client.](https://github.com/honojs/hono/issues/2719)
```
yusukebe, line 5: I think the following sentence is not proper here:

> At this time, responses returned from middleware are [not inferred by the client.](https://github.com/honojs/hono/issues/2719)

How about this change?

```diff
diff --git a/docs/guides/rpc.md b/docs/guides/rpc.md
index b83e8fe..c8bc211 100644
--- a/docs/guides/rpc.md
+++ b/docs/guides/rpc.md
@@ -2,7 +2,10 @@

 The RPC feature allows sharing of the API specifications between the server and the client.

-Using the `AppType` exported from your server code, the Hono Client can infer both the input type(s) specified by the Validator, and the output type(s) emitted by handlers returning `c.json()`. At this time, responses returned from middleware are [not inferred by the client.](https://github.com/honojs/hono/issues/2719)
+Using the `AppType` exported from your server code, the Hono Client can infer both the input type(s) specified by the Validator, and the output type(s) emitted by handlers returning `c.json()`.
+
+> [!NOTE]
+> At this time, responses returned from middleware are [not inferred by the client.](https://github.com/honojs/hono/issues/2719)

 > [!NOTE]
 > For the RPC types to work properly in a monorepo, in both the Client's and Server's tsconfig.json files, set `"strict": true` in `compilerOptions`. [Read more.](https://github.com/honojs/hono/issues/2270#issuecomment-2143745118)
```

yusukebe, line 5: I think you should explain what is `AppType`. What do you think of it?

yusukebe:
Hi @ambergristle 

These are good improvements! I've left comments. Thanks!

docs/guides/rpc.md
ambergristle: agreed! it's definitely more of a note/call-out

ambergristle: agreed. i was going back + forth, but it is a little unclear

ambergristle: does this work?
```suggestion
First, export the `typeof` your Hono app (commonly called `AppType`)—or just the routes you want available to the client—from your server code.

By accepting `AppType` as a generic parameter, the Hono Client can infer both the input type(s) specified by the Validator, and the output type(s) emitted by handlers returning `c.json()`.

> [!NOTE]
> At this time, responses returned from middleware are [not inferred by the client.](https://github.com/honojs/hono/issues/2719)
```

ambergristle:
thanks for the feedback @yusukebe!

i've updated accordingly. please let me know if you'd like me to make any further changes

docs/guides/rpc.md
yusukebe: Looks good!

yusukebe (approved):
LGTM!

yusukebe:
@ambergristle 

Your work will help many people who read the docs. Thank you for the contribution!

PR #608, opened 2025-03-06
docs: document `addValidatedData` method

ambergristle:
### TL;DR
Added some documentation on the `HonoRequest.addValidatedData` method, along with some helpful guidance on validating request data with `hono-openapi`.

I've also made a few other additions/tweaks to provide some additional clarity on validation. Please let me know if those are unhelpful.

_I'm not sure how headings get translated to the table of contents on the right side of the docs UI. Will H3s get read in?_

### Guides/Validation
A user on Discord identified a valid use-case for overriding (validated) request data in middleware. At the same time, the `addValidatedData` method can have unexpected consequences when used without the proper typing. I tried to strike a balance between documenting the method and mis-encouraging its use.
- https://discord.com/channels/1011308539819597844/1347182247337529415/1347229968400257046

_Should I also add an example of how to use the method correctly? It seemed out of scope to me, but I welcome your thoughts!_

### Examples/HonoOpenAPI
A user on Discord ran into some issues using scalar to render openapi docs generated by `hono-openapi`. The issue proved to be an unnecessary specification of the request body in `describeRoute`.
- https://discord.com/channels/1011308539819597844/1343725111601270824/1347232975372419166

### DIFF
--- a/docs/guides/validation.md
+++ b/docs/guides/validation.md
@@ -30,12 +30,13 @@ app.post(
       body: body,
     }
   }),
-  //...
+  // ...
 ```
 
 Within the handler you can get the validated value with `c.req.valid('form')`.
 
 ```ts
+// after your middleware...
 , (c) => {
   const { body } = c.req.valid('form')
   // ... do something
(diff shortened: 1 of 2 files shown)


ambergristle:
thanks for identifying these solutions @MathurAditya724!

let me know if there's anything i missed, or if a different direction would be better

yusukebe:
@ambergristle 

`HonoRequest.addValidatedData` is a public method, but it's not for end-user.

PR #612, opened 2025-03-15
fix(docs): Add note about .tsx extension requirement for JSX.

luka-hash:
Just a minor info to JSX page.


yusukebe (approved):
LGTM!

yusukebe:
Hi @luka-hash 

This will be a super help for users. Thank you so much!

PR #613, opened 2025-03-18
Tweaked wording in Node.js documentation

mjomble:
I've never heard of HTTP(S) requests being referred to as "access".

Also, "If access comes in and returns a file, write this code" sounds like the "access" is already returning a file, before you've written any code. I rephrased it as "and you want to return a file", so we're talking about the developer's intent, not something that is already happening.


yusukebe (approved):
LGTM!

PR #614, opened 2025-03-19
Update middleware.md

basuke:
Trivial count typo with plural form fix.

### DIFF
--- a/docs/guides/middleware.md
+++ b/docs/guides/middleware.md
@@ -26,7 +26,7 @@ If the handler returns `Response`, it will be used for the end-user, and stoppin
 app.post('/posts', (c) => c.text('Created!', 201))
 ```
 
-In this case, four middleware are processed before dispatching like this:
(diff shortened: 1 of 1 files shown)


docs/guides/middleware.md
```diff
@@ -26,7 +26,7 @@ If the handler returns `Response`, it will be used for the end-user, and stoppin
 app.post('/posts', (c) => c.text('Created!', 201))
 ```
 
-In this case, four middleware are processed before dispatching like this:
+In this case, three middlewares are processed before dispatching like this:
```
yusukebe, line 29: We use `middleware` not `middlewares` because it's an uncountable noun.

basuke: Understood. Thanks for clarifying!

yusukebe: @basuke Thanks!

But, I think `four` should be `three`.

PR #615, opened 2025-03-20
chore(docs): Clerk is using hono

panteliselef:
Greetings from the Clerk team 👋. We are happy to share that we are using Hono + CF workers extensively. 

We'd like to be included in the usages section.


yusukebe:
@panteliselef 

Cool! It's an honor you use Hono! Thank you!

PR #619, opened 2025-03-21
Update cookie examples

Mael-Abgrall:
The documentation around cookies is a little bit confusing, it took me a while to understand where each parameter goes, and where are actually the secrets.

This PR aims to make the doc a bit less fun, but more straightforward

### DIFF
--- a/docs/helpers/cookie.md
+++ b/docs/helpers/cookie.md
@@ -17,35 +17,26 @@ import {
 
 ## Usage
 
+### Regular cookies
+
+```ts
+setCookie(c, 'cookie_name', 'cookie_value')
+const yummyCookie = getCookie(c, 'cookie_name')
+deleteCookie(c, 'cookie_name')
+const allCookies = getCookie(c)
+```
+### Signed cookies
+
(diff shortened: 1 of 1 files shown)


docs/helpers/cookie.md
```diff
@@ -17,35 +17,26 @@ import {
 
 ## Usage
 
+### Regular cookies
+
+```ts
+setCookie(c, 'cookie_name', 'cookie_value')
```
yusukebe, line 23: I think you should write these code inside of a handler like the following. Otherwise, some users may not understand what is `c`.

```ts
app.get('/cookie', (c) => {
  setCookie(c, 'cookie_name', 'cookie_value')
  const yummyCookie = getCookie(c, 'cookie_name')
  deleteCookie(c, 'cookie_name')
  const allCookies = getCookie(c)
  // ...
})
```

Mael-Abgrall: done!

yusukebe (approved):
LGTM!

yusukebe:
@Mael-Abgrall 

Thanks! Great job!

PR #621, opened 2025-03-25
fix: added missing export keyword

MusheghMov:
Added missing export keyword for the "Lambda response streaming" section. AWS Lambda function will throw an error if the handler is not exported


yusukebe (approved):
LGTM!

yusukebe:
@MusheghMov 

Thank you! Merging.

PR #622, opened 2025-03-25
fix(docs): modify import statement to type-only imports explicitly

Enchan1207:
## summary

Modified `import` to [*Type-Only Imports*](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export) explicitly.

```ts
// from
import { AppType } from '.'

// to
import type { AppType } from '.'
```

## details

In most cases, its difference is not a problem.

### conditions that cause problems

If compiler option [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax) is enabled, type import must be written as follows:

```ts
// A. NG
import { AppType } from '.'

// B. OK
import type { AppType } from '.'

// C. it is also OK
import { type AppType } from '.'
```

However, in pattern C, **server-side code may be mixed into client-side** .

### why this happen

It is caused by rewriting rule of TypeScript.

The statement `import { type AppType } from '@/module'` is rewritten into `import {} from '@/module'`.
It means `@/module` will still be imported as a value after compilation.

I created the repository ([Enchan1207/hono_rpc_import_style_differences](https://github.com/Enchan1207/hono_rpc_import_style_differences)) for reproducing its behavior.

The option `verbatimModuleSyntax` is not enabled as default, so it might be nitpick.

### DIFF
--- a/docs/guides/rpc.md
+++ b/docs/guides/rpc.md
@@ -50,7 +50,7 @@ export type AppType = typeof route
 On the Client side, import `hc` and `AppType` first.
 
 ```ts
-import { AppType } from '.'
+import type { AppType } from '.'
 import { hc } from 'hono/client'
 ```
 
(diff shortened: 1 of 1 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@Enchan1207 

Thanks! It should only import the types. I'll merge this.

PR #630, opened 2025-04-04
docs: use `html` helper instead of dangerouslySetInnerHTML

WilliamEspegren:
is there a reason why dangerouslySetInnerHTML is used here?

I think the formatting and neatness is improved by using the html helper

### DIFF
--- a/docs/helpers/websocket.md
+++ b/docs/helpers/websocket.md
@@ -140,6 +140,7 @@ ws.addEventListener('open', () => {
 ```tsx
 import { Hono } from 'hono'
 import { createBunWebSocket } from 'hono/bun'
+import { html } from 'hono/html'
 
 const { upgradeWebSocket, websocket } = createBunWebSocket()
 
--- a/docs/helpers/websocket.md
+++ b/docs/helpers/websocket.md
@@ -153,17 +153,15 @@ app.get('/', (c) => {
       </head>
       <body>
         <div id='now-time'></div>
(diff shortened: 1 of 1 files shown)


docs/helpers/websocket.md
```diff
@@ -153,17 +153,15 @@ app.get('/', (c) => {
...
-            __html: `
-        const ws = new WebSocket('ws://localhost:3000/ws')
-        const $nowTime = document.getElementById('now-time')
-        ws.onmessage = (event) => {
-          $nowTime.textContent = event.data
-        }
-        `,
-          }}
-        ></script>
+        {html`
```
yusukebe, line 156: Please add the import statement for `hono/html` at the top: 

```ts
import { Hono } from 'hono'
import { createBunWebSocket } from 'hono/bun'
import { html } from 'hono/html'
```

yusukebe:
@WilliamEspegren 

> is there a reason why dangerouslySetInnerHTML is used here?

Exactly, not reason. We can use `hono/html`. I've left a comment.

docs/helpers/websocket.md
WilliamEspegren: Ahh Ofc, great catch

WilliamEspegren:
There we go @yusukebe :)

yusukebe:
@WilliamEspegren 

Thanks! Merging.

PR #633, opened 2025-04-08
fix search button styling

futantan:
Fix the search button styling on mobile, here is how it looks like:
CleanShot 2025-04-08 at 22 25 35@2x


yusukebe (approved):
LGTM!

PR #634, opened 2025-04-09
Docs: Add note about Cloudflare's RPC preference over fetch for Durable Objects

braden-w:
## Description

The current Hono documentation for Cloudflare Durable Objects shows examples using the `fetch` handler pattern. However, as of compatibility date `2024-04-03`, Cloudflare recommends using RPC methods over the `fetch` handler pattern for Durable Objects.

While Hono still works perfectly with the `fetch` handler pattern, it would be helpful to:
1. Add a note about this Cloudflare recommendation
2. Potentially provide an example using both patterns

## Current Documentation
The current example at https://hono.dev/examples/cloudflare-durable-objects shows only the `fetch` handler pattern:

```typescript
export class Counter {
  value: number = 0
  state: DurableObjectState
  app: Hono = new Hono()

  constructor(state: DurableObjectState) {
    // ...
  }

  async fetch(request: Request) {
    return this.app.fetch(request)
  }
}
```

## Suggested Addition

Consider adding a note like:

> **Note:** As of Cloudflare Workers compatibility date `2024-04-03`, it's recommended to use RPC methods instead of the `fetch` handler for Durable Objects. While Hono continues to work with the `fetch` handler pattern shown above, you might want to consider the RPC pattern for new projects. See [Cloudflare's documentation](https://developers.cloudflare.com/durable-objects/best-practices/create-durable-object-stubs-and-send-requests/) for more details.

## Additional Context
- Reference: https://developers.cloudflare.com/durable-objects/best-practices/create-durable-object-stubs-and-send-requests/
- This change only affects projects with compatibility dates >= 2024-04-03
- Both patterns remain functional, but RPC is now considered best practice by Cloudflare

## Impact
This documentation update would help developers make informed decisions about which pattern to use in their Hono applications with Durable Objects.

### DIFF
--- a/examples/cloudflare-durable-objects.md
+++ b/examples/cloudflare-durable-objects.md
@@ -2,6 +2,8 @@
 
 By using Hono, you can write [Durable Objects](https://developers.cloudflare.com/durable-objects/) application easily.
 
+> **Note:** As of Cloudflare Workers compatibility date `2024-04-03`, it's recommended to use RPC methods instead of the `fetch` handler for Durable Objects. While the example below using Hono's `fetch` handler pattern continues to work, consider using the RPC pattern for new projects. See [Cloudflare's documentation](https://developers.cloudflare.com/durable-objects/best-practices/create-durable-object-stubs-and-send-requests/) for more details.
+
 Hono can handle a `fetch` event of Durable Objects and you can use it with the powerful router.
 
 ```ts


yusukebe:
Hi @braden-w 

Thank you for the PR!

In my opinion. This is just a document, so I think it's fine to just include the method that we recommend for users. I don't think it's necessary to include the `fetch` pattern, as it's not recommended and is verbose. Instead, how about including an example of the RPC pattern, though it will be simple?

yusukebe:
> Sounds great! We could give an example of the RPC pattern

Thanks you.

> but it probably wouldn't require using Hono at all 😅 I'll think of an example!

I think using the RPC in Hono makes a lot of sense!

braden-w:
@yusukebe Just pushed an updated example for the docs! Let me know if it looks good

examples/cloudflare-durable-objects.md
```diff
@@ -1,48 +1,146 @@
...
+  // ... Counter class implementation stays the same ...
+}
+
+// Create a new Hono app
+const app = new Hono();
+
+// Add routes to interact with the Durable Object
+app.get('/counter', async (c) => {
+  const env = c.env;
+  const id: DurableObjectId = env.COUNTER.idFromName("counter");
```
yusukebe, line 119: If you pass the generics for Bindings property:

```ts
type Bindings = {
  COUNTER: DurableObjectNamespace<Counter>
}

const app = new Hono<{ Bindings: Bindings }>()
```

You don't need to specify the type `id: DurableObjectId`:

```ts
app.get('/counter', async (c) => {
  // ...
  const id = env.COUNTER.idFromName('counter') // `DurableObjectId `
  // ...
})
```

examples/cloudflare-durable-objects.md
```diff
@@ -1,48 +1,146 @@
...
-    })
-
-    this.app.get('/', async (c) => {
-      return c.text(this.value.toString())
-    })
+You can use Hono as the router in your Cloudflare Worker, calling RPC (Remote Procedure Calls) to interact with [Durable Objects](https://developers.cloudflare.com/durable-objects/). This is the recommended approach as of Cloudflare Workers compatibility date `2024-04-03`.
+
+## Example: Counter Durable Object
+
+### Initial Setup
```
yusukebe, line 7: I think you don't need to put the document for general implementation not using Hono since the users can know it from the Cloudflare document. And I don't want to use `create-cloudflare` as much as possible. If the command menus are changed, we have to update the docs.

This means you can remove the `## Example: Counter Durable Object` section and add just an example using Hono with `Counter`.

yusukebe:
Hi @braden-w 

Thank you for updating. I've left the comments!

braden-w:
Sounds great and agree with changes! Will make them later tonight

braden-w:
Just made another push! Let me know what you think

yusukebe (approved):
LGTM!

braden-w:
Awesome, happy to help! And thank you for responding so fast 🤯

PR #641, opened 2025-04-14
Docs: Clarify routing priority earlier

kevinsimper:
Moved the "Routing priority" section up and expanded it to better explain how Hono matches routes based on registration order. This aims to proactively address common confusion where parameterized routes might unintentionally capture requests intended for more specific static routes defined later. Added examples to illustrate the importance of ordering static vs. dynamic paths and also clarified the grouping order example.

### DIFF
--- a/docs/api/routing.md
+++ b/docs/api/routing.md
@@ -63,6 +63,90 @@ app.get('/posts/:id/comment/:comment_id', async (c) => {
 })
 ```
 
+## Routing Priority
+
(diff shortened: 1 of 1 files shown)


docs/api/routing.md
```diff
@@ -63,6 +63,90 @@ app.get('/posts/:id/comment/:comment_id', async (c) => {
 })
 ```
 
+## Routing Priority
```
yusukebe, line 66: The description is perfect. It's clear!

But I think you should not move this section to the above. The current position below of "Routing with host Header value" is good. This "Routing Priority" is important, but the user should read it after understanding basic routing syntaxes. It's my opinion.

CleanShot 2025-04-15 at 10 51 18@2x

PR #655, opened 2025-04-29
Update aws-lambda.md

perpil:
+ Add esbuild as a dev dependency so docker doesn't need to be running for `cdk deploy`
+ Remove APIG, having both a lambda function url and APIG is unnecessary
+ Emit function url endpoint
+ Update node to 22

### DIFF
--- a/docs/getting-started/aws-lambda.md
+++ b/docs/getting-started/aws-lambda.md
@@ -20,6 +20,7 @@ mkdir my-app
 cd my-app
 cdk init app -l typescript
 npm i hono
+npm i -D esbuild
 mkdir lambda
 touch lambda/index.ts
 ```
@@ -29,6 +30,7 @@ mkdir my-app
 cd my-app
 cdk init app -l typescript
 yarn add hono
+yarn add -D hono
 mkdir lambda
 touch lambda/index.ts
 ```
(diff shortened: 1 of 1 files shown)


docs/getting-started/aws-lambda.md
```diff
@@ -20,6 +20,7 @@ mkdir my-app
 cd my-app
 cdk init app -l typescript
 npm i hono
+npm i -D esbuild
```
yusukebe, line 23: I'm not super familiar with AWS Lambda, but I wonder if `esbuid` is needed. The script in `package.json` created by CDK uses `tsc` to build an app.

perpil: A little background on this, if you are using NodeJsFunction, either docker running locally or esbuild is required to bundle.  See [here](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda_nodejs-readme.html#local-bundling)

When I followed the instructions as is, I got:

```
❯ cdk deploy
ERROR: Cannot connect to the Docker daemon
```

Also see this [SO post](https://stackoverflow.com/a/74312067/3006039)

yusukebe: @perpil 

Understood well. Thank you for the explanation!

yusukebe (approved):
LGTM!

PR #660, opened 2025-05-06
Nodejs/gracefull shutdown

Kanahiro:
This PR adds sample codes to handle process/server close event, based on https://github.com/honojs/node-server/issues/86

### DIFF
--- a/docs/getting-started/nodejs.md
+++ b/docs/getting-started/nodejs.md
@@ -80,7 +80,17 @@ import { Hono } from 'hono'
 const app = new Hono()
 app.get('/', (c) => c.text('Hello Node.js!'))
 
-serve(app)
+const server = serve(app)
+
+// gracefully shutdown
+process.on("SIGINT", () => {
+  console.log("Ctrl-C was pressed")
+  server.close()
+})
+process.on("SIGTERM", () => {
+  console.log("SIGTERM received")
+  server.close()
+})
 ```
 
 ## 3. Run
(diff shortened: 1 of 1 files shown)


docs/getting-started/nodejs.md
```diff
@@ -80,7 +80,22 @@ import { Hono } from 'hono'
 const app = new Hono()
 app.get('/', (c) => c.text('Hello Node.js!'))
 
-serve(app)
```
yusukebe, line 83: Showing how to do a graceful shutdown is okay, but first, we want to show the shortest code. So I think it's better to leave it unchanged and add a new code section like the following patch:

```diff
diff --git a/docs/getting-started/nodejs.md b/docs/getting-started/nodejs.md
index 2564e55..5537d15 100644
--- a/docs/getting-started/nodejs.md
+++ b/docs/getting-started/nodejs.md
@@ -80,14 +80,21 @@ import { Hono } from 'hono'
 const app = new Hono()
 app.get('/', (c) => c.text('Hello Node.js!'))

+serve(app)
+```
+
+If you want to gracefully shut down the server, write it like this:
+
+```ts
 const server = serve(app)

 // graceful shutdown
-process.on("SIGINT", () => {
+process.on('SIGINT', () => {
   server.close()
   process.exit(0)
 })
-process.on("SIGTERM", () => {
+
+process.on('SIGTERM', () => {
   server.close((err) => {
     if (err) {
       console.error(err)
```

Kanahiro: Thanks, fixed.

yusukebe (approved):
LGTM!

yusukebe:
@Kanahiro 

Thanks! Merging.

PR #661, opened 2025-05-06
Update cloudflare-pages.md - Add missing basePath method

Chinoman10:
- Add missing 'basePath' in the example provided.

The example clearly states the typical use-case of having the functions under the `/api` route, however if you do that, you'll need to add `.basePath("/api");` to the `new Hono` instruction.


yusukebe (approved):
LGTM!

yusukebe:
@Chinoman10 

You are right. Thanks!

PR #663, opened 2025-05-14
Improve alt text.

Afterlifepro:
This pull request adds proper alt text to several images, while removing alt text from all images where the alt text is simply the filename/other nondescript values. 

This should make it easier to find when alt text was omitted.

I currently don't have the time to write alt text for images, but this should be a good base.

It should also be noted that it is reccomended to avoid images which just show text, for example: the code block in /index.md or the benchmarks in /docs/concepts/benchmarks.md

### DIFF
--- a/docs/concepts/benchmarks.md
+++ b/docs/concepts/benchmarks.md
@@ -92,41 +92,41 @@ Let's see the results.
 
 The following screenshots show the results on Node.js.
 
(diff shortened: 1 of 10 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@Afterlifepro 

Awesome. These alts are not proper, as you said. Thanks!

PR #665, opened 2025-05-16
docs: Add wasi support

### LINKED ISSUE #664: Docs on WebAssembly + WASI support
Thanks to the flexibility and Web Standards-forward stance of Hono and WebAssembly + WASI (and hard work by many contributors), Hono actually runs *completely unmodified* on WebAssembly workloads powered by [StarlingMonkey](https://github.com/bytecodealliance/StarlingMonkey) 🎉 .

We're in the midst of adding [an official Hono example project in `jco`](https://github.com/bytecodealliance/jco/pull/699), and would love to have the compatibility echoed here as well -- WebAssembly + WASI is a new platform that Hono supports completely out of the box.

If there's appetite, I'd love to add some documentation that explains how to get started/how to use Hono from server-side (and also browser side!) WebAssembly, thanks to `jco` and related Wasm/JS ecosystem projects.

vados-cosmonic:
This PR adds documentation around using Hono with WebAssembly + WASI via `jco`, `componentize-js`, and StarlingMonkey (the JS Wasm ecosystem stack).

Resolves #664

### DIFF
--- a/docs/getting-started/webassembly-wasi.md
+++ b/docs/getting-started/webassembly-wasi.md
@@ -1,0 +1,283 @@
+# WebAssembly (w/ WASI)
+
+[WebAssembly][wasm-core] is a secure, sandboxed, portable runtime that runs inside and outside browsers. In practice,
+many languages *compile to* WebAssembly, and WebAssembly runtimes *run* WebAssembly binaries (`.wasm`).
+
+While core WebAssembly has *no* access to things like the local filesystem or sockets, the [WebAssembly System Interface][wasi]
+steps in to enable defining a platform under WebAssebly workloads. This means that *with* WASI, WebAssembly can operate on
+files, sockets, and much more.
+
+::: info
+Want to peek at the WASI interface yourself? check out [`wasi:http`][wasi-http]
+:::
+
+Support for WebAssembly w/ WASI in JS is powered underneat by [StarlingMonkey][sm], and thanks to the focus on Web standards in
+both StarlingMonkey and Hono, **Hono works *out of the box with WASI-enabled WebAssembly ecosystems.**
+
+[sm]: https://github.com/bytecodealliance/StarlingMonkey
+[wasm-core]: https://webassembly.org/
+[wasi]: https://wasi.dev/
+[bca]: https://bytecodealliance.org/
+
+## 1. Setup
+
+The WebAssembly JS ecosystem provides tooling to make it easy to get started building WASI-enabled WebAssembly components:
+
+- [StarlingMonkey][sm] is a WebAssembly-aware JS runtime
+- [`componentize-js`][componentize-js] is a tool for building WebAssembly components from Javascript files
+- [`jco`][jco] is a multi-tool for building, generating types, and running components in environments like NodeJS or the browser
+
+::: info
+Webassembly has an open ecosystem is an open source, with core projects stewarded primarily by the [Bytecode Alliance][bca] and it's members.
+
+New features, issues, pull requests and other types of contributions are always welcome.
+:::
+
+While a starter for WebAssembly Hono is not yet available, you can start a WebAssembly Hono project just
+like any other:
+
+::: code-group
+
+```sh [npm]
+mkdir my-app
+cd my-app
+npm init
+npm i hono
+npm i -D @bytecodealliance/jco @bytecodealliance/componentize-js rollup
+npm i -D rollup @rollup/plugin-typescript @rollup/plugin-node-resolve
+```
+
+```sh [yarn]
+mkdir my-app
+cd my-app
+npm init
+yarn add hono
+yarn add -D @bytecodealliance/jco @bytecodealliance/componentize-js
+yarn add -D rollup @rollup/plugin-typescript @rollup/plugin-node-resolve
+```
+
+```sh [pnpm]
+mkdir my-app
+cd my-app
+npm init
+pnpm add hono
+pnpm add -D @bytecodealliance/jco @bytecodealliance/componentize-js
+pnpm add -D rollup @rollup/plugin-typescript @rollup/plugin-node-resolve
+```
+
+```sh [bun]
+mkdir my-app
+cd my-app
+npm init
+bun add hono
+bun add -D @bytecodealliance/jco @bytecodealliance/componentize-js
+```
+
+:::
+
+Move into `my-app` and install the dependencies.
+
+::: code-group
+
+```sh [npm]
+cd my-app
+npm i
+```
+
+```sh [yarn]
+cd my-app
+yarn
+```
+
+```sh [pnpm]
+cd my-app
+pnpm i
+```
+
+```sh [bun]
+cd my-app
+bun i
+```
+
+:::
+
+Since `componentize-js` (and `jco` which re-uses that functionality) does not (yet) work with multiple JS files,
+bundling is necessary, so [`rollup`][rollup] is used for package managers that do not already bundle.
+
+The following Rollup configuration (`rollup.config.mjs`) should also be used:
+
+```js
+import typescript from "@rollup/plugin-typescript";
+import resolve from "@rollup/plugin-node-resolve";
+import dotenv from "rollup-plugin-dotenv";
+import { nodeResolve } from "@rollup/plugin-node-resolve";
+
+export default {
+  input: "component.ts",
+  external: /wasi:.*/,
+  output: {
+    file: "dist/component.js",
+    format: "esm",
+  },
+  plugins: [
+    dotenv(),
+    resolve(),
+    typescript({ noEmitOnError: true }),
+    nodeResolve(),
+  ],
+};
+```
+
+[jco]: https://github.com/bytecodealliance/jco
+[componentize-js]: https://github.com/bytecodealliance/componentize-js
+
+## 2. Hello World
+
+If you use the App Router, Edit `app/api/[[...route]]/route.ts`. Refer to the [Supported HTTP Methods](https://nextjs.org/docs/app/building-your-application/routing/route-handlers#supported-http-methods) section for more options.
+
+```ts
+import { Hono } from 'hono'
+
+const app = new Hono();
+
+app.get('/hello', (c) => {
+  return c.json({
+    message: 'Hello WASM!',
+  })
+})
+
+export const GET = handle(app)
+export const POST = handle(app)
+
+app.fire();
+```
+
+## 3. Build
+
+
+First, transpile your Typescript into javsacript:
+
+::: code-group
+
+```sh [npm]
+npx tsc -p .
+```
+
+```sh [yarn]
+yarn tsc -p .
+```
+
+```sh [pnpm]
+pnpx tsc -p .
+```
+
+```sh [bun]
+bun build --target=bun --outfile=dist/component.js ./component.ts
+```
+
+:::
+
+Then, bundle your transpile javascript into a single JS file, with it's dependencies:
+
+::: code-group
+
+```sh [npm]
+npx rollup -c
+```
+
+```sh [yarn]
+yarn rollup -c
+```
+
+```sh [pnpm]
+pnpx rollup -c
+```
+
+```sh [bun]
+bun build --target=bun --outfile=dist/component.js ./component.ts
+```
+
+:::
+
+
+To build your WebAssembly component, use `jco` (and indirectly `componentize-js`):
+
+::: code-group
+
+```sh [npm]
+npx jco componentize -w wit -o dist/component.wasm dist/component.js
+```
+
+```sh [yarn]
+yarn jco componentize -w wit -o dist/component.wasm dist/component.js
+```
+
+```sh [pnpm]
+pnpx jco componentize -w wit -o dist/component.wasm dist/component.js
+```
+
+```sh [bun]
+bun run jco componentize -w wit -o dist/component.wasm dist/component.js
+```
+
+:::
+
+
+## 3. Run
+
+To run your Hono WebAssembly HTTP server, you can use any WASI-enabled WebAssembly runtime:
+
+- [`wasmtime`][wasmtime]
+- `jco` (runs in NodeJS)
+
+In this guide, we'll use `jco serve` since it's already installed.
+
+::: warn
+`jco serve` is meant for development, and is not recommended for production use.
+:::
+
+[wasmtime]: https://wasmtime.dev
+
+::: code-group
+
+```sh [npm]
+npx jco serve dist/component.wasm
+```
+
+```sh [yarn]
+yarn jco serve dist/component.wasm
+```
+
+```sh [pnpm]
+pnpx jco serve dist/component.wasm
+```
+
+```sh [bun]
+bun run jco serve dist/component.wasm
+```
+
+:::
+
+You should see output like the following:
+
+```
+$ npx jco serve dist/component.wasm
+Server listening @ localhost:8000...
+```
+
+Sending a request to `localhost:8000/hello` will produce the JSON output you've specified in your Hono application.
+
+
+## More information
+
+For more information on the WASI, WebAssembly components and more, see the following resources:
+
+- [BytecodeAlliance Component Model book][cm-book]
+- [`jco` codebase][jco]
+  - [`jco` example components][jco-example-components] (in particular the [Hono example][jco-example-component-hono])
+- [`componentize-js` codebase][compnentize-js]
+
+[cm-book]: https://component-model.bytecodealliance.org/
+
+[jco-example-components]: https://github.com/bytecodealliance/jco/tree/main/examples/components
+[jco-example-component-hono]: https://github.com/bytecodealliance/jco/tree/main/examples/components/http-server-hono
--- a/docs/getting-started/webassembly-wasi.md
+++ b/docs/getting-started/webassembly-wasi.md
@@ -53,7 +53,7 @@ cd my-app
 npm init
 yarn add hono
 yarn add -D @bytecodealliance/jco @bytecodealliance/componentize-js
-yarn add -D rollup @rollup/plugin-typescript @rollup/plugin-node-resolve
+yarn add -D rollup @rollup/plugin-typescript @rollup/plugin-node-resolve tslib
 ```
 
 ```sh [pnpm]
@@ -62,7 +62,7 @@ cd my-app
 npm init
 pnpm add hono
 pnpm add -D @bytecodealliance/jco @bytecodealliance/componentize-js
-pnpm add -D rollup @rollup/plugin-typescript @rollup/plugin-node-resolve
+pnpm add -D rollup @rollup/plugin-typescript @rollup/plugin-node-resolve tslib
 ```
 
 ```sh [bun]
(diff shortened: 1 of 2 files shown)


docs/getting-started/webassembly-wasi.md
```diff
@@ -0,0 +1,340 @@
...
+```sh [bun]
+bun run jco serve dist/component.wasm
+```
+
+:::
+
+You should see output like the following:
+
+```
+$ npx jco serve dist/component.wasm
```
yusukebe, line 316: I tried it with Bun. But, the following message has been displayed, and `localhost:8000` has hung.

CleanShot 2025-05-17 at 04 42 15@2x

docs/getting-started/webassembly-wasi.md
```diff
@@ -0,0 +1,340 @@
...
+## 3. Run
+
+To run your Hono WebAssembly HTTP server, you can use any WASI-enabled WebAssembly runtime:
+
+- [`wasmtime`][wasmtime]
+- `jco` (runs in NodeJS)
+
+In this guide, we'll use `jco serve` since it's already installed.
+
+::: warn
```
yusukebe, line 287: This should be `::: warning`.

docs/getting-started/webassembly-wasi.md
yusukebe, line 1: Can you also update `.vitepress/config.ts` to add the menu on the left bar?

```diff
diff --git a/.vitepress/config.ts b/.vitepress/config.ts
index 1da6358..3838c61 100644
--- a/.vitepress/config.ts
+++ b/.vitepress/config.ts
@@ -69,6 +69,10 @@ const sidebars = (): DefaultTheme.SidebarItem[] => [
         text: 'Ali Function Compute',
         link: '/docs/getting-started/ali-function-compute',
       },
+      {
+        text: 'WebAssembly',
+        link: '/docs/getting-started/webassembly-wasi',
+      },
       {
         text: 'Service Worker',
         link: '/docs/getting-started/service-worker',
```

docs/getting-started/webassembly-wasi.md
vados-cosmonic: Thanks, fixed!

docs/getting-started/webassembly-wasi.md
vados-cosmonic: Thanks, done!

docs/getting-started/webassembly-wasi.md
vados-cosmonic: Thanks for noting this -- I see this consistently on every *second* request.

I think this *might* be a bug in how [Wizer](https://github.com/bytecodealliance/wizer) is processing the WebAssembly component (there is setup code that runs once, and it *seems* like the handler is being attached twice?). However, since NodeJS handles it well (and building for `--target=node` produces no change), it's possible the bug is on Bun's side.

Should I remove the Bun support for now and instead leave a message explaining the situation? Once the underlying problems are fixed, the Bun instructions could be added back in.

yusukebe: @vados-cosmonic 

It shows the same result if I build that with `npx rollup`:

CleanShot 2025-05-19 at 15 35 42@2x

vados-cosmonic: Ah, so if you run with `wasmtime serve` this actually *doesn't* show up:

```console
wasmtime serve -S cli,http dist/component.wasm
```

This is a problem specific to `jco serve`, not actually the component/implementation! I'll look into this since I maintain `jco`.

yusukebe: Ah, great! `wasmtime serve` works fine.

Yes. I'll wait for your response.

vados-cosmonic: Thank you for your patience :bow:

docs/getting-started/webassembly-wasi.md
```diff
@@ -0,0 +1,340 @@
...
+```sh [bun]
+bun run jco serve dist/component.wasm
+```
+
+:::
+
+You should see output like the following:
+
+```
+$ npx jco serve dist/component.wasm
```
yusukebe, line 316: Does `jco` really exist on npm now? It returns 404.

CleanShot 2025-06-26 at 17 28 55@2x

docs/getting-started/webassembly-wasi.md
```diff
@@ -0,0 +1,340 @@
...
+ *
+ * Since both Hono and StarlingMonkey are aligned Web Standards (WinterCG/WinterTC),
+ * this enables Hono to run smoothly in WASI-enabled (`wasi:http`) Webassembly environments.
+ *
+ * See: https://github.com/bytecodealliance/ComponentizeJS#using-starlingmonkeys-fetch-event
+ * See: https://hono.dev/docs/concepts/web-standard
+ * See: https://wintertc.org/
+ * See: https://github.com/WebAssembly/wasi-http
+ */
+app.fire();
```
yusukebe, line 221: Now `app.fire()` is deprecated:

* https://github.com/honojs/hono/pull/4231
* https://github.com/honojs/hono/issues/3127

So, use `fire` function in `hono/service-worker`.

```ts
import { fire } from 'hono/service-worker'

// ...

fire(app)
```

docs/getting-started/webassembly-wasi.md
```diff
@@ -0,0 +1,340 @@
...
+## 2. Retrieve WIT dependencies
+
+[`wasi:http`][wasi-http] is the standardized interface for dealing with HTTP requests (whether it's receiving them or sending them out),
+and since our component reuses that functionality, we have to declare it in the [WIT world][wit-world] our component exports:
+
+[wit-world]: https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#wit-worlds
+
+First, let's set up the component's WIT world:
+
+```wit
```
yusukebe, line 153: Unfortunately, syntax for `wit` is not supported in VitePress:

CleanShot 2025-06-26 at 17 33 28@2x

So, can you set it to `txt`? If it supports `wit`, we will be able to turn on its `wit`.

vados-cosmonic: Ah, that's unfortunate, thanks for pointing that out, will switch to `txt`!

docs/getting-started/webassembly-wasi.md
rajsite: > I'd like to wait for the WinterTC-forward approach to get fixed and contributed to StarlingMonkey.

@vados-cosmonic is there an issue or PR tracking that behavior? Think I was running into it as well at some point.

vados-cosmonic: Hey Raj, good point! This actually came up in the last JS community meeting and we talked about it personally there. There isn't an issue filed yet on StarlingMonkey but the team is very aware of the problem (there's already a fast-but-not-complete fix floating around).

vados-cosmonic: Yeah unfortunately it's not fixed yet -- I will update when it is!

There is another option here for how to fix it which involves using a Hono adapter w/ `wasi:http`, so I may pursue that option because it would be quicker (and works today).

rajsite: > There is another option here for how to fix it which involves using a Hono adapter w/ wasi:http, so I may pursue that option because it would be quicker (and works today). I'll send an update as well if I do that!

I feel like that would involve a substantial amount of code or additional libraries (likely a custom hono adapter for wasi web components). Even in it's basic usage you need to be aware of fairly tricky concepts / concepts most JS developers generally do not need to deal with like resource management with `Symbol.dispose`.

I'd really hope we can lean towards standard fetch-event based APIs as the primary source instead in these docs and have a fix for the upstream issues. I think waiting for that would be beneficial for the most users 🙏.

rajsite: I also think in practice most usages are of components today would be via single run contexts in wasmtime like envs (wasmtime directly, wasmcloud, spin, etc) and users would be successful with what is there today. The jco behavior of [reusing instancing can be reasonably argued as not the common / expected pattern at least today and not recommended for production](https://github.com/bytecodealliance/jco/issues/535).

I'd advocate in the short term to update the docs to describe testing with wasmtime and avoid `jco serve` in the docs until `jco serve` behavior is updated as the linked issue describes or the upstream StarlingMonkey issue is addressed.

vados-cosmonic: > I feel like that would involve a [non-trivial amount of code](https://github.com/wasmCloud/typescript/pull/207/commits/6807c4f58c314a5995254bd864f1a6b412dc58e9#diff-53edbc5bba06d9e7248ab7e8019ed8eb772cc8a1bdf4779ddedfae274acb5960L1-L93) or additional libraries (likely a custom hono adapter for wasi web components). Even in it's basic usage you need to be aware of fairly tricky concepts / concepts most JS developers generally do not need to deal with like resource management with Symbol.dispose.

Yup, it would require a non-trivial amount of code, but the average user wouldn't have to write that, they'd be able to "just" pull it in, like they do for any other Hono adapters today! Adapters are meant to hide that stuff from users, so that's what the adapter here would do as well.

I think I wasn't clear -- the intent was to create that adapter *upstream* and make it easily `npm install`-able (i.e. part of upstream `jco`) and then it would "just work" for most use cases.

> I'd really hope we can lean towards standard fetch-event based APIs as the primary source instead in these docs and have a fix for the upstream issues. I think waiting for that would be beneficial for the most users 🙏.

We certainly should! I agree here, and I certainly hope it's fixed quickly upstream -- we already explored a quick solution so it's more like someone just needs to do the work and make the final solution robust.

That said, I'd remind you that fetch-event based APIs are not actually yet a standard -- we're building towards the future here (which I agree with), but using technology that works today is important. There is no reason we can't support both, just like 99% of servers today are not yet on fetch-event based APIs :)

Having something that is not broken is better than having only having something that is broken.

> I also think in practice most usages are of components today would be via single run contexts in wasmtime like envs (wasmtime directly, wasmcloud, spin, etc) and users would be successful with what is there today. The jco behavior of https://github.com/bytecodealliance/jco/issues/535.

Yep, this is absolutely true, and something we discussed on the call this week -- we actually already have a plan for jco request isolation (we'll likely be using VMs on the node side), but that feature is independent of proper support for reusing a component instance. 

Disposing wasm instances after use is certainly recommended, but it is not the *only* way to use components, and would be quite limiting -- the ecosystem should support both, which is why this is worth fixing in StarlingMonkey *in the first place*. If that goal wasn't reasonable, this would be a purely `jco` bug.

> I'd advocate in the short term to update the docs to describe testing with wasmtime serve and avoid jco serve in the docs until jco serve behavior is updated as the linked jco issue describes or the upstream StarlingMonkey issue is addressed.

While `jco serve` is not meant for production, it's pretty clearly noted in the documentation that it's not meant for production (I added those docs myself) -- I'd certainly be open to adding an explicit warning in when it's run to make sure it's abundantly clear. 

That said, requiring a beginner following the guide to download and install `wasmtime` I think takes them out of flow for a basic example setup. Not every interface must be production ready, and `jco serve` is useful because it is quick to experiment with and does not require external tools.

rajsite: > Yup, it would require a non-trivial amount of code, but the average user wouldn't have to write that, they'd be able to "just" pull it in, like they do for any other Hono adapters today! Adapters are meant to hide that stuff from users, so that's what the adapter here would do as well.

I think my concern is creating a hono adapter library intending to adapt wasi:http to the JS fetch api surface that is standalone implemented in the hono adapter repo. I'd hope a library like that has thorough testing and seems like a useful candidate library to be implemented outside of hono and used here (related to the discussion of a std lib that maps wasi:filesystem <-> nodejs filesystem api this would be having a wasi:http/incoming-handler <-> fetch-event, and I assume you're not advocating for it here but such a library could also have wasi:http/outgoing-handler <-> fetch). The std library could help insulate projects like hono from directly having to handle concerns of wasi api surfaces (like unusual JS imports containing versioning and type generation).

I'm also concerned about increasing the volume of JS code in user mode, particularly with incomplete testing. Debugging in the JS component context is challenging and leveraging a mode that seems partially tested (instance re-use for multiple requests) may warrant extra test coverage as well.

As an aside, I'm not against the idea of an adapter at all. I'd just hope the goal is to minimize the surface area of wasi apis exposed directly in hono vs through a shared tested library implementing common interfaces. Adapter would certainly be helpful for env, filesystem, etc.

Would you agree that once the upstream bug in fetch-event is addressed that the extra code for directly using wasi:http/incoming-handler would be redundant and removed from a hono adapter in favor of fetch-event?

> Having something that is not broken is better than having only having something that is broken.

I disagree with the assessment that what is here is broken. I have manually validated a fairly decent size chunk of features of hono using the fetch-event api (and have submitted some minor patches to get some things working). I'd ask that those are re-evaluated with the custom wasi:http implementation to avoid regressions:
- simple:  https://github.com/rajsite/wasm-component-examples/blob/main/examples/hono-app/src/lib/app.ts
- oauth, oidc, jsx, proxy: https://github.com/rajsite/wasm-component-examples/blob/main/examples/hono-features-app/src/lib/app.ts
- openapi: https://github.com/rajsite/wasm-component-examples/blob/main/examples/hono-openapi-app/src/lib/app.ts

> That said, requiring a beginner following the guide to download and install wasmtime I think takes them out of flow for a basic example setup.

I would expect that to be a short-term requirement. The desire for this extra work makes me wonder if this is expected to be a long-term issue. Are there discussions where I can get a better understanding of the complexity that would warrant the extra work and testing in the short-term?

vados-cosmonic: > I think my concern is creating a hono adapter library intending to adapt wasi:http to the JS fetch api surface that is standalone implemented in the hono adapter repo. I'd hope a library like that has thorough testing and seems like a useful candidate library to be implemented outside of hono and used here (related to the discussion of a std lib that maps wasi:filesystem <-> nodejs filesystem api this would be having a wasi:http/incoming-handler <-> fetch-event, and I assume you're not advocating for it here but such a library could also have wasi:http/outgoing-handler <-> fetch). The std library could help insulate projects like hono from directly having to handle concerns of wasi api surfaces (like unusual JS imports containing versioning, p1 style callback based handlers, type generation, etc).

Yup that is definitely the intent -- it's not that the adapter is implemented *inside* hono in any way, like you said it would be implemented outside and used here!

> I'm also concerned about increasing the volume of JS code in user mode, particularly with incomplete testing. Debugging in the JS component context is challenging and leveraging a mode that seems partially tested (instance re-use for multiple requests) may warrant extra test coverage as well.

That's a great observation -- I agree it may be challenging but the right fix here is to improve debugging, and improving the tools, not shying away from this use case. Debugging is hard, yes, but this is something that we area absolutely in control of making better.

If we expect WebAssembly to be truly widely adopted, having great coverage around valid use cases (component reuse is not invalid) is IMO an improvement.

> As an aside, I'm not against the idea of an adapter at all. I'd just hope the goal is to minimize the surface area of wasi apis exposed directly in hono vs through a shared tested library implementing common interfaces. Adapter would certainly be helpful for env, filesystem, etc.
>
> Would you agree that once the upstream bug in fetch-event is addressed that the extra code for directly using wasi:http/incoming-handler would be redundant and removed from a hono adapter in favor of fetch-event?

I *fully and emphatically* agree here! The intended WASI API exposure of what I'm proposing is *~zero* -- it's an import of the adapter (IMO this an alternative "external" `fire` function). Allowing Hono to run on new platforms is a [well established thing](https://github.com/honojs/node-server), and that flexibility is a boon. 

You're hitting on an important point here -- `fetch-event` *isn't enough* to build a high-quality adapter for Hono on the WebAssembly platform. Config, FS, and other things that are exposed is important to provider in a Hono-native way. This is like building a Cloudflare adapter -- as long as it Just Works (tm), ~no one spends much time debugging Cloudflare's adapter.

In a world where fetch-event is supported, we certainly wouldn't *need* the `wasi:http` adapter per say. That said, as `wasi:http` is a lower level construct, there are advanced things that you might be able to do with it that are not yet included in WinterTC standards just yet. An admittedly contrived example is *trailer support* ([not yet supported](https://github.com/whatwg/fetch/issues/981), AFAIK).

There is no need to have *only one* way to do things in this arena -- just like we do not expect everyone to switch to fetch-event and never respond to requests another way again. In fact, this isn't even the only adapter to consider -- `wasi:sockets` exists! All of these approaches have different trade offs.

> I disagree with the assessment that what is here is broken. I have manually validated a fairly decent size chunk of features of hono using the fetch-event api with wasmtime and wasmcloud (and have submitted some minor patches to get some things working). I'd ask that those are re-evaluated with the custom wasi:http implementation to avoid regressions:

Thanks for suggesting this! It would be great to have these. If you wouldn't mind contributing these examples upstream to Jco, We can cover them under test for every commit going forward (and it'd serve as a great example!). We can continue the discussion there because I think that'd be a great place to have the tests regardless.

I would note though, that if you're using those platforms with your current setup you wouldn't be affected by these changes -- you're not *forced* to use a `wasi:http` adapter, if it did exist.

> I would expect that to be a short-term requirement. The desire for this extra work makes me wonder if this is expected to be a long-term issue. Are there discussions where I can get a better understanding of the complexity that would warrant the extra work and testing in the short-term?

Well, the "short term" often ends up being longer than we expect -- I thought this PR would take at max a couple hours but here we are. I can't force anyone on StarlingMonkey to prioritize the fetch-event fix ahead of the work they're *already* doing, so I can't depend on how fast the issue might get fixed.

I think the best place to keep up with the discussion would be [in the StarlingMonkey repo](https://github.com/bytecodealliance/StarlingMonkey/issues/247) -- I think that if we expect fetch-event to be the only way *or* if StarlingMonkey should not support reusable components (that's totally valid, and worth discussing), that would be the right place to discuss.

As far as I see it -- Hono is a platform and WASI support for it does not mea WinterTC support -- it means full access to the suite of WASI standards that are available in the platform. In addition to this example just plain not working right now, having an adapter is a quick way to more robust (working) support for requests and even more functionality.

Happy to rip out the manual `wasi:http` adapter code for fetch-event when it makes sense!

rajsite: > There is no need to have only one way to do things in this arena -- just like we do not expect everyone to switch to fetch-event and never respond to requests another way again.

I'm not saying that. Hono itself is aligned on wintercg and [Web standards](https://hono.dev/docs/concepts/web-standard#web-standards). If devs dig into the source of the adapters I think they'd be more comfortable seeing the fetch-event source. As you described, today there isn't a compelling behavior for wasi:http usage over fetch-event usage and if there is tomorrow would make sense as the default exposed to users in docs!

> Happy to rip out the manual wasi:http adapter code for fetch-event when it makes sense!

Sounds great! The best code to maintain is no code!

vados-cosmonic:
Hey @yusukebe it's been a while, but this is finally ready for another review, I think! :bow: yoroshiku!

docs/getting-started/webassembly-wasi.md
```diff
@@ -0,0 +1,350 @@
...
+  external: /wasi:.*/,
+  output: {
+    file: 'dist/component.js',
+    format: 'esm',
+  },
+})
+```
+
+::: info
+Feel free to use any other bundlers that you're more comfortable with (`rolldown`, `esbuild`, `rolldown`, etc)
```
yusukebe: `rolldown` is duplicated?

yusukebe:
@vados-cosmonic 

Cool! It works well on my machine. I've left the nitpick, but others are great. I'll merge after you fix it!

yusukebe (approved):
LGTM!

PR #666, opened 2025-05-16
Update request.md with examples parsing with same names

kevinsimper:
Just making it more clear it is not just for files but also for input fields that has the same name :)


yusukebe (approved):
LGTM!

yusukebe:
@kevinsimper 

Thanks!

PR #667, opened 2025-05-20
Update routing.md with regex explanations

kevinsimper:
A few examples of how to best approach the regex routing :)

### DIFF
--- a/docs/api/routing.md
+++ b/docs/api/routing.md
@@ -75,13 +75,20 @@ app.get('/api/animal/:type?', (c) => c.text('Animal!'))
 
 ## Regexp
 
+You can apply a Regular Expression to a path parameter. The syntax is `/:paramName{regexp}`.
+
 ```ts twoslash
 import { Hono } from 'hono'
 const app = new Hono()
 // ---cut---
+// Example: :date must be a sequence of digits ([0-9]+),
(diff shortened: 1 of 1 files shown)


docs/api/routing.md
```diff
@@ -92,8 +99,17 @@ app.get('/post/:date{[0-9]+}/:title{[a-z]+}', async (c) => {
...
-app.get('/posts/:filename{.+\\.png}', async (c) => {
-  //...
+// Example: :filename must end with .png and have a non-empty name part.
+// The regex `.+\\.png$` ensures this:
+//   - `.+` matches one or more leading characters (the filename base).
+//   - `\\.png$` matches the literal ".png" string at the end of the segment.
+//
+// This means paths like /file/dog.png will match, making filename "data.png".
+// Paths like /file/.png (empty name part) or /file/data.png.txt (wrong suffix) will not match.
+app.get('/post/:filename{.+\\.png$}', async (c) => {
```
yusukebe, line 109: It may not be clear that `filename` can contain `/`. For example, make the app as follows:

```ts
app.get('/file/:filename{.+\\.png$}', async (c) => {
  const { filename } = c.req.param()
  // ...
})
```

And explain that accessing `/file/foo/bar.png` will make `filename` be `foo/bar.png`. How about it?

PR #669, opened 2025-05-21
Update better-auth.md

aswnss-m:
### Summary

This PR simplifies the `AuthType` definition and updates the way we type the Hono app.

### What Changed

* Changed `AuthType` from `{ Variables: { session, user } }` to just `{ session, user }`.
* Updated Hono app creation to use `Hono<AuthType>()` instead of `Hono<{ Bindings: ..., Variables: AuthType }>()`.

### Why

The previous typing structure prevented TypeScript from correctly inferring types when using `c.set("session", session)` and `c.get("session")`. This refactor ensures that session and user values are properly typed and accessible within the context.

### DIFF
--- a/examples/better-auth.md
+++ b/examples/better-auth.md
@@ -60,10 +60,8 @@ export const auth = betterAuth({
 })
 
 export type AuthType = {
-  Variables: {
     user: typeof auth.$Infer.Session.user | null
     session: typeof auth.$Infer.Session.session | null
-  }
 }
 ```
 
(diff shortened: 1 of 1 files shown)


aswnss-m:
Also note that this change matches the example integration code sample from the [better-auth docs](https://www.better-auth.com/docs/integrations/hono)

yusukebe (approved):
LGTM!

yusukebe:
@aswnss-m Thanks!

PR #671, opened 2025-05-25
Docs(example): add better-auth-on-cloudflare

bytaesu:
## Example: better-auth-on-cloudflare

> I’ve closed the previous PR and opened a new one with the same changes, but with a cleaner commit history.

An example Hono application demonstrating Better Auth integration, optimized for deployment on Cloudflare Workers.


yusukebe (approved):
LGTM!

PR #674, opened 2025-05-30
Add cap checkpoint to third-party middlewares

tiagorangel1:
Adds Cap's checkpoint to third-party middlewares list


yusukebe:
@tiagorangel1 

I didn't know the Cap, but interesting. Thanks! Merging.

PR #677, opened 2025-06-14
docs: Add Stytch to Authentication middlewares list

max-stytch:


yusukebe:
@max-stytch 

Thanks! Merging.

PR #678, opened 2025-06-15
docs(stytch-auth): Add Integration example with Stytch Authentication

max-stytch:
This PR adds a more-complete guide to using the new `@hono/stytch-auth` package with the Stytch FE SDKs. It is intended to be a partner to the example app created in https://github.com/honojs/examples/pull/228


yusukebe:
@max-stytch 

This also looks good to me. I think we can merge this right now. But if you have any fixes to do, please create a new PR. Thanks!

PR #686, opened 2025-06-17
Update swagger-ui.md

francelwebdev:
following the documentation, the /ui route returns an error. so I added what was missing.


yusukebe:
@francelwebdev Thanks!

PR #688, opened 2025-06-18
docs: add theme-color meta tag

McAllaster:
Add a [theme-color meta tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/theme-color) with Hono's orange branding.

I personally think the docs site looks slick with the orange accent on compatible devices, but feel free to reject this if you feel otherwise.

### DIFF
--- a/.vitepress/config.ts
+++ b/.vitepress/config.ts
@@ -395,6 +395,7 @@ export default defineConfig({
         content: 'https://hono.dev/images/hono-title.png',
       },
     ],
+    ['meta', { name: 'theme-color', content: '#E36002' }],
     ['meta', { property: 'og:type', content: 'website' }],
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @McAllaster  

Interesting! But I don't know if we should merge.  

On my iPhone, I turned on the dark mode on iOS and the webpage dark mode, which currently looks like the following:  

IMG_6676  

I applied this change:  

IMG_6675  

I think the orange is incompatible.  

But it's a good idea. I want to hear others' opinions.

McAllaster:
Yeah, Hono's orange is definitely a bold and fiery color! If you're not sure about it, I'm sure other folks feel the same. Keeping it as-is is definitely the safer choice, but I figured I'd at least put this out for discussion.

PR #697, opened 2025-07-20
docs: fix undefined type in cloudflare-durable-objects.md

lotap:
Add `<Env = unknown>` to Counter declaration so that `Env` isn't undefined on the constructor line below.

I encountered that after copy & pasting the example and running the cf-typegen script (`wrangler types --env-interface CloudflareBindings`)

### DIFF
--- a/examples/cloudflare-durable-objects.md
+++ b/examples/cloudflare-durable-objects.md
@@ -15,7 +15,7 @@ You can use Hono as the router in your Cloudflare Worker, calling RPCs (Remote P
 import { DurableObject } from 'cloudflare:workers'
 import { Hono } from 'hono'
 
-export class Counter extends DurableObject {
+export class Counter<Env = unknown> extends DurableObject {
   // In-memory state
   value = 0
 


examples/cloudflare-durable-objects.md
```diff
@@ -15,7 +15,7 @@ You can use Hono as the router in your Cloudflare Worker, calling RPCs (Remote P
 import { DurableObject } from 'cloudflare:workers'
 import { Hono } from 'hono'
 
-export class Counter extends DurableObject {
+export class Counter<Env = unknown> extends DurableObject {
```
yusukebe, line 18: Simply, you should do this:

```diff
diff --git a/examples/cloudflare-durable-objects.md b/examples/cloudflare-durable-objects.md
index 8920b86..4820ad8 100644
--- a/examples/cloudflare-durable-objects.md
+++ b/examples/cloudflare-durable-objects.md
@@ -19,7 +19,7 @@ export class Counter extends DurableObject {
   // In-memory state
   value = 0

-  constructor(ctx: DurableObjectState, env: Env) {
+  constructor(ctx: DurableObjectState, env: unknown) {
     super(ctx, env)

     // `blockConcurrencyWhile()` ensures no requests are delivered until initialization completes.
```

lotap: Thanks for the suggestion! I've updated the PR to reflect that change.

I also realized `constructor(ctx: DurableObject["ctx"], env: DurableObject["env"])` would preserve inheritance from `DurablbeObject` if it ever changes.

Let me know if you prefer that syntax and I will update again.

yusukebe (approved):
LGTM!

yusukebe:
@lotap 

There are many ways to add types in the TypeScript world, but this PR is okay to keep simple. Thank you!

PR #700, opened 2025-07-23
Uses performance.now instead of Date.now

### LINKED ISSUE #699: Should the docs use `performance.now` for performance example?
In the [middleware concepts](https://github.com/honojs/website/blob/52b9c7e96f651fcd18ad13c91a338f7f024395c5/docs/concepts/middleware.md?plain=1#L16) example, the code shows:  
```js
app.use(async (c, next) => {
  const start = Date.now()
  await next()
  const end = Date.now()
  c.res.headers.set('X-Response-Time', `${end - start}`)
})
```

Should this example use the `performance.now()` call instead?  
See https://developer.mozilla.org/en-US/docs/Web/API/Performance/now  
> "Unlike [Date.now](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now), the timestamps returned by performance.now() are not limited to one-millisecond resolution. Instead, they represent times as floating-point numbers with up to microsecond precision."

> "(...) `performance.now()` is coarsened based on whether or not the document is [cross-origin isolated](https://developer.mozilla.org/en-US/docs/Web/API/Window/crossOriginIsolated)."

I'm happy to start a PR.

DiogoNeves:
This PR fixes #699 

The PR replaces the use of `Date.now()` with `performance.now()` (see issue for details)

### DIFF
--- a/docs/concepts/middleware.md
+++ b/docs/concepts/middleware.md
@@ -13,9 +13,9 @@ import { Hono } from 'hono'
 const app = new Hono()
 // ---cut---
 app.use(async (c, next) => {
-  const start = Date.now()
+  const start = performance.now()
   await next()
-  const end = Date.now()
+  const end = performance.now()
   c.res.headers.set('X-Response-Time', `${end - start}`)
 })
 ```


yusukebe (approved):
LGTM!

yusukebe:
Hi @DiogoNeves 

Great idea! Thank you so much!

DiogoNeves:
Thanks 🙏

PR #646, opened 2025-04-21
docs: delete @scalar/hono-api-reference

ysknsid25:
resolves https://github.com/honojs/website/issues/645

Issue #645, opened 2025-04-21
@scalar/hono-api-reference has already been deprecated


ysknsid25:
about [Hono OpenAPI > Serve API Docs](https://hono.dev/examples/hono-openapi#%F0%9F%8C%90-serve-api-docs).

version `0.8.2` of `@scalar/hono-api-reference`,  `apiReference` has already been deprecated.

![Image](https://github.com/user-attachments/assets/3e5f23de-80c5-46a3-89c7-8b4dc6816deb)

![Image](https://github.com/user-attachments/assets/a4aa4463-c649-4f2f-a89d-249590c8e374)

So should I delete the documentation? Or should I write that it is deprecated?🤔

yusukebe:
Hi @ysknsid25 

I'm not super familiar with `@scalar/hono-api-reference`, but if it's `deprecated`, you can remove that section. I think it's unnecessary to continue to show the feature that will be obsolete. Thanks.

mrlubos:
Wait a second @ysknsid25 @yusukebe, put Scalar back 😂 Clearly only the `apiReference` export has been deprecated, not the whole package. The correct usage is as follows, you can see the official [documentation](https://guides.scalar.com/scalar/scalar-api-references/integrations/hono).

```ts
import { Hono } from 'hono'
import { Scalar } from '@scalar/hono-api-reference'

const app = new Hono()

// Use the middleware to serve the Scalar API Reference at /scalar
app.get('/scalar', Scalar({ url: '/doc' }))

// Or with dynamic configuration
app.get('/scalar', Scalar((c) => {
  return {
    url: '/doc',
    proxyUrl: c.env.ENVIRONMENT === 'development' ? 'https://proxy.scalar.com' : undefined,
  }
}))

export default app

```

PR #701, opened 2025-07-24
docs: add Scalar back to examples

mrlubos:
Related https://github.com/honojs/website/issues/645
Related https://github.com/honojs/website/pull/646

### DIFF
--- a/.vitepress/config.ts
+++ b/.vitepress/config.ts
@@ -273,12 +273,16 @@ export const sidebarsExamples = (): DefaultTheme.SidebarItem[] => [
         link: '/examples/zod-openapi',
       },
       {
-        text: 'Swagger UI',
-        link: '/examples/swagger-ui',
-      },
-      {
         text: 'Hono OpenAPI',
         link: '/examples/hono-openapi',
+      },
+      {
+        text: 'Scalar',
+        link: '/examples/scalar',
+      },
+      {
+        text: 'Swagger UI',
+        link: '/examples/swagger-ui',
       },
     ],
   },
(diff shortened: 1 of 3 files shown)


.vitepress/config.ts
mrlubos, line 1: I moved Hono OpenAPI up because it needs to be configured before Scalar so it wouldn't make sense to have Scalar first in the menu

.vitepress/config.ts
```diff
@@ -272,14 +272,18 @@ export const sidebarsExamples = (): DefaultTheme.SidebarItem[] => [
...
-      {
-        text: 'Swagger UI',
-        link: '/examples/swagger-ui',
-      },
       {
         text: 'Hono OpenAPI',
         link: '/examples/hono-openapi',
       },
+      {
+        text: 'Scalar',
```
yusukebe, line 280: Can you put the `Scalar` under `Swagger UI`? Since `Swagger UI` was created before `Scalar`, I would like to place it above.

mrlubos: Sure. I had them ordered by quality – Scalar is a vastly superior product, rapidly gaining adoption among companies including the likes of Microsoft. That being said, I moved it below Swagger UI.

yusukebe: Thank you for understanding.

yusukebe (approved):
LGTM!

yusukebe:
@mrlubos 

> would you be open to adding a guide on how to generate SDKs?

Yes. Sounds good.

PR #704, opened 2025-07-30
Update Vercel documentation.

anthonyshew:
## Hono is now zero-configuration on Vercel

Hi! We've updated our support for Hono so that it can be used without any configuration required. These docs can now be updated for better DX.

Thanks!


package.json
yusukebe, line 1: Can you remove this change? `package.json` should not be updated in this PR.

yusukebe:
@anthonyshew 

The zero-configuration is beautiful. Thanks.

PR #718, opened 2025-08-11
docs(cloudflare-workers): update load env section

ryuapp:
- Add information about `process.env`
- Add that `.env` files can also be used


yusukebe:
@ryuapp Thanks!

PR #719, opened 2025-08-11
docs: remove service worker mode section

ryuapp:
`app.fire()` and Service Worker mode has already been deprecated, so we can remove the section.

https://developers.cloudflare.com/workers/reference/migrate-to-module-workers/#migrate-a-worker


yusukebe:
@ryuapp 

Nice catch. Thanks!

PR #720, opened 2025-08-11
docs(fastly): use `fire()` from `hono/service-worker` instead of `app.fire()`

ryuapp:
Related to https://github.com/honojs/starter/pull/102


yusukebe (approved):
LGTM!

PR #729, opened 2025-09-01
ci: add formatting and build checks to workflow

3w36zj6:
Since there were many cases of forgetting to run format, I've added a simple CI workflow that checks both formatting and the build.

Following recent discussions in honojs/hono and honojs/middleware, Bun and `.tool-versions` have been introduced for CI purposes.


bun.lock
yusukebe, line 1: We have not added any lockfiles, but now it's good to add `bun.lock`.

yusukebe (approved):

PR #730, opened 2025-09-01
ci: add autofix workflow for automated code formatting

3w36zj6:
cf. https://github.com/honojs/website/pull/729#issuecomment-3241639356


3w36zj6:
It seems to be fine.

PR #735, opened 2025-09-05
docs(rpc): fix type reference in client example

karur4n:
## Summary
Fixed incorrect type reference in the RPC documentation's client example.

## Changes
- Changed `typeof routes` to `typeof app` in the Client type definition to match the correct variable name used in the example

The example was using `app` variable but the type was referencing non-existent `routes` variable, which would cause TypeScript errors when users copy this code.


yusukebe (approved):
LGTM!

PR #737, opened 2025-09-08
docs(jwt): Add details for verifyOptions

buckett:
The JWT middleware verifies the timestamp of the JWT (if present) and the issuer (if included in the verifyOptions). There is no longer the ability to supply the `issuer` directly on the options.


yusukebe:
@buckett 

Looks good! Thank you.

PR #739, opened 2025-09-13
fix: updated hono-openapi docs for v1

MathurAditya724:
- [x] Updated the examples page

### DIFF
--- a/examples/hono-openapi.md
+++ b/examples/hono-openapi.md
@@ -39,7 +39,7 @@ import { Hono } from 'hono'
 import {
   describeRoute,
   resolver,
-  validator as vValidator,
+  validator,
 } from 'hono-openapi'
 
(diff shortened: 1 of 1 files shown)


examples/hono-openapi.md
```diff
@@ -46,12 +36,11 @@ Use `describeRoute` for route documentation and validation:
 
 ```ts
 import { Hono } from 'hono'
-import { describeRoute } from 'hono-openapi'
-// You can import these for your preferred validation library
 import {
+  describeRoute,
   resolver,
   validator as vValidator,
```
yusukebe, line 42: Just curious. Should we rename `validator` to `vValidator`. I think `validator` is okay because it does not depend on a specific validator.

yusukebe (approved):
LGTM!

PR #744, opened 2025-09-18
docs: refactor import to use AppType for client typing

MrUprizing:
Docs: refactor import to use AppType for client typing


yusukebe:
Hi @MrUprizing 

It's the same thing passing  `typeof app` to `hc` from creating `AppType` and passing it. So, I think this change is unnecessary.

yusukebe:
@MrUprizing 

Thank you for your explanation. These reasons are meaningful. Let's go with this PR!

PR #765, opened 2025-10-25
docs(websocket): update Bun example to use direct imports

braden-w:
The Bun with JSX example in the WebSocket Helper documentation was using the deprecated `createBunWebSocket()` pattern. This updates the example to import `upgradeWebSocket` and `websocket` directly from `hono/bun`, which is the current recommended approach.

The deprecated pattern looked like this:

```tsx
import { createBunWebSocket } from 'hono/bun'
const { upgradeWebSocket, websocket } = createBunWebSocket()
```

The updated pattern now imports directly:

```tsx
import { upgradeWebSocket, websocket } from 'hono/bun'
```

This aligns with the deprecation notice in the source code and matches the pattern already shown in the Import section of the same documentation page.


docs/guides/validation.md
yusukebe: Can you remove this change to `validation.md`? It's not related to this PR.

braden-w:
@yusukebe Just changed! Thanks for the suggestion.

PR #770, opened 2025-11-04
docs: add Cloudflare R2 TUS Upload example

riderx:
This Pr adds a exemple of implementing TUS protocole with Hono and clouflare workers, r2 and durable object


examples/cloudflare-r2-tus-upload.md
```diff
@@ -0,0 +1,213 @@
...
+
+Or with Bun:
+
+```bash
+bun add @capgo/hono-r2-tus-uploader
+```
+
+## Configuration
+
+First, configure your R2 bucket and Durable Object bindings in `wrangler.toml`:
```
yusukebe: How about adding a `wrangler.jsonc` example?

PR #772, opened 2025-11-08
Revise Cloudflare Workers deployment guide

The-LukeZ:
Updated instructions for setting up Cloudflare Workers deployment with GitHub Actions, including token management and workflow configuration.

The old workflow example was very likely going to fail because of missing nodejs and package manager setup - also if you didn't set your `account_id` in the `wrangler.jsonc/toml`.

### DIFF
--- a/docs/getting-started/cloudflare-workers.md
+++ b/docs/getting-started/cloudflare-workers.md
@@ -303,8 +303,8 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v4
         with:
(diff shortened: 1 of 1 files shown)


The-LukeZ:
Yep, I did that. See yourself: https://github.com/The-LukeZ/upvote-engine

I don't know if you can see the action tab but the first workflow run used the exact example of the workflow.

yusukebe:
@The-LukeZ 

I think only the change in #778 is needed. We can use #778 instead of this PR.

PR #775, opened 2025-11-10
docs: Clarify RPC param/query typing, Middleware behavior

ambergristle:
### Motivations
- there is some confusion around the need to stringify RPC param and query values, and then parse/coerce them server-side: https://github.com/honojs/middleware/issues/707#issuecomment-3475812615
- the docs suggest that responses can't be returned from middleware: https://discord.com/channels/1011308539819597844/1012485912409690122/1436017379334160485
  - auth, rate limiting, or other middleware may return a `Response` to exit early (rather than throwing)
  - is there any reason not to return a `Response` from middleware?

### What's Changed
- updated the RPC guide to include the type + conversion requirements
- modified the Middleware guide to discuss early returns

### DIFF
--- a/docs/guides/rpc.md
+++ b/docs/guides/rpc.md
@@ -255,7 +255,7 @@ const res = await client.posts[':id'].$get({
     id: '123',
   },
   query: {
-    page: '1', // `string`, even though server expects `number`
+    page: '1', // `string`, converted by the validator to `number`
   },
 })
 ```


yusukebe (approved):
LGTM!

yusukebe:
Hi @ambergristle 

Awesome! The docs became really clear. Thank you for your contribution!

PR #781, opened 2025-11-16
docs: update Vercel documentation link to point to the correct backend section

shubhxho:


yusukebe:
@shubhxho Nice catch! Thank you!

PR #854, opened 2026-04-21
docs(context): reword ContextVariableMap description and add warning (#853)

### LINKED ISSUE #853: "ContextVariableMap" description in API-Context section is misleading
I was writing a custom auth middleware, that validates user session and then adds user ID to context. I wrote the following thing:

```typescript
export const auth = createMiddleware(async (c, next) => {
  const token = getCookie(c, 'token');

  // other logic...
  const decoded = decodeJwt(token);
  c.set('userId', decoded.userId);

  await next();
});
```

As TypeScript can't infer, that "userId" was set after this middleware was used, I read a documentation about Hono's Context and saw that you can extend "ContextVariableMap" to, as it clearly states, "add type definitions to variables when a specific middleware is used" by doing:

```typescript
declare module 'hono' {
  interface ContextVariableMap {
    result: string
  }
}
```
This will indeed add a variable to context, but not "when a specific middleware is used", instead this will just add it to global context regardless of whether middleware was used or not, therefore polluting global context and creating a possibility of runtime errors due to incorrect type inference. I will give an example so you can see how this can be a problem.

```typescript
// authMiddleware.ts
// this is our custom middleware

declare module 'hono' {
  interface ContextVariableMap {
    userId: string
  }
}

export const auth = createMiddleware(async (c, next) => {
  const token = getCookie(c, 'token');

  // other logic...
  const decoded = decodeJwt(token);
  c.set('userId', decoded.userId);

  await next();
});
```

```typescript
// router.ts

import { auth } from './authMiddleware' 

// uses custom middleware, that adds a variable to context
const router1 = new Hono().use('*', auth);

router1.get('/', async (c) => {
  c.get('userId') // no error, variable should be present after middleware ran, as expected
});

// doesn't use custom middleware, variable should not be accessible
const router2 = new Hono()

router2.get('/', async (c) => {
  c.get('userId') // also no error, but the variable will be undefined, as auth middleware didn't run
}
```
What should be done instead is passing a generic type to "createMiddleware" factory and then chaining middleware during router initialization:

```typescript
// authMiddleware.ts

// now we add a type for Context Variables
type AuthEnv = {
  Variables: {
    userId: string;
  };
}

// and pass it here
export const auth = createMiddleware<AuthEnv>(async (c, next) => {
  const token = getCookie(c, 'token');

  // other logic...
  const decoded = decodeJwt(token);
  c.set('userId', decoded.userId);

  await next();
});
```

```typescript
// router.ts

import { auth } from './authMiddleware' 

// uses custom middleware
const router1 = new Hono().use('*', auth);

router1.get('/', async (c) => {
  c.get('userId') // no error, as expected
});

// doesn't use custom middleware, variable should not be accessible
const router2 = new Hono()

router2.get('/', async (c) => {
  c.get('userId') // this will now raise an error as there is no such variable
}
```

I'm not saying ContextVariableMap is useless or should be deprecated, but the description is misleading and adding variables in this way will add them to global context, making them accessible anywhere without proper type inference. I think the description should be reworded and/or redirect to "Extending the Context in Middleware" for this specific case: adding variables only after a specific middleware ran, not adding them globally and doing unnecessary type guards.

I can and would like to take on this issue, if you'd prefer delegating it to someone and not doing it yourself.

incandesc3nce:
Resolves #853 

As requested by @yusukebe in [this comment](https://github.com/honojs/website/issues/853#issuecomment-4286689067), I reworded ContextVariableMap introductory description to clarify its purpose and use cases and added a warning about the global nature of ContextVariableMap and encouraged using generic type arguments in createMiddleware.

Please let me know if anything should be changed.

### DIFF
--- a/docs/api/context.md
+++ b/docs/api/context.md
@@ -482,7 +482,7 @@ const mw = createMiddleware(async (c, next) => {
 
 const app = new Hono()
 
-// handler uses the middleware 
+// handler uses the middleware
 app.get('/foo', mw, (c) => {
   const val = c.get('result') // ✅ val is a string and typed as such, as expected
 })
--- a/docs/api/context.md
+++ b/docs/api/context.md
@@ -461,7 +461,43 @@ app.use(async (c, next) => {
 
 ## ContextVariableMap
 
-For instance, if you wish to add type definitions to variables when a specific middleware is used, you can extend `ContextVariableMap`. For example:
+::: warning
+In most cases you should prefer passing a generic type argument to `createMiddleware`, as it scopes the variable types to only the routes where the middleware is explicitly used. See [Extending the Context in Middleware](/docs/guides/middleware#extending-the-context-in-middleware) and [Type Inference Across Chained Middleware](/docs/guides/middleware#type-inference-across-chained-middleware).
+
+`ContextVariableMap` adds types **globally** to all contexts, regardless of whether the middleware that sets the variable has actually run. This means `c.get('result')` will appear type-safe even in handlers where your middleware was never registered, potentially hiding `undefined` bugs at runtime.
+
+Take a look at the following example:
+
+```ts
+declare module 'hono' {
+  interface ContextVariableMap {
+    result: string
+  }
+}
+
+const mw = createMiddleware(async (c, next) => {
+  c.set('result', 'some values')
+  await next()
+})
+
+const app = new Hono()
+
+// handler uses the middleware 
+app.get('/foo', mw, (c) => {
+  const val = c.get('result') // ✅ val is a string and typed as such, as expected
+})
+
+// handler doesn't use the middleware
+app.get('/bar', (c) => {
+  const val = c.get('result') // ❌ val is undefined but typed as a string, which can lead to runtime errors
+})
+```
+
+:::
+
+You can augment the `ContextVariableMap` interface to define types for context variables globally across your entire application. This is appropriate when a variable is set by middleware that is applied app-wide and is guaranteed to exist in the context.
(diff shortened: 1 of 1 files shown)


docs/api/context.md
```diff
@@ -461,7 +461,43 @@ app.use(async (c, next) => {
 
 ## ContextVariableMap
 
-For instance, if you wish to add type definitions to variables when a specific middleware is used, you can extend `ContextVariableMap`. For example:
+::: warning
+In most cases you should prefer passing a generic type argument to `createMiddleware`, as it scopes the variable types to only the routes where the middleware is explicitly used. See [Extending the Context in Middleware](/docs/guides/middleware#extending-the-context-in-middleware) and [Type Inference Across Chained Middleware](/docs/guides/middleware#type-inference-across-chained-middleware).
```
yusukebe: I think you can remove this.

The problem is not only for `createMiddleware`, so simply, I think it's good to remove it. What do you think of this?

incandesc3nce: I agree with you, the problem isn't limited to `createMiddleware` and I think it's better to just highlight the underlying issue.

Overall, I think the updated description is fine as it is and the warning should stay and specify this pitfall, but without the part about `createMiddleware`.

Should everything else stay the same? I will update the PR once confirmed or any other changes are requested.

yusukebe: Yes. The others are okay!

yusukebe (approved):
LGTM!

yusukebe:
@incandesc3nce Thank you!

PR #858, opened 2026-05-04
docs(nextjs): use `getRequestListener` instead of vercel helper

ryuapp:
Since `@hono/node-server/vercel` was removed in v2, we've switched to using `getRequestListener` as an alternative.
https://github.com/honojs/node-server/releases/tag/v2.0.0


yusukebe (approved):
LGTM!

yusukebe:
@ryuapp 

Nice! I forgot it. Thanks!

PR #860, opened 2026-05-07
docs(stacks): use type-only import for AppType

o8n:
Same as #622 and #744, applied to docs/concepts/stacks.md.

Both `AppType` imports in this page are only used as a type parameter to `hc<>()`, so they should be type-only imports. With `verbatimModuleSyntax: true`, the current example fails to compile with `TS1484`.

<details>
<summary>Reproduction</summary>

Node.js v22.12.0, TypeScript 6.0.3, Hono 4.12.18.

`tsconfig.json`

```json
{
  "compilerOptions": {
    "target": "ESNext",
    "module": "ESNext",
    "moduleResolution": "Bundler",
    "strict": true,
    "verbatimModuleSyntax": true,
    "skipLibCheck": true,
    "noEmit": true
  }
}
```

`tsc --noEmit` on the current example in `docs/concepts/stacks.md`

```
client.ts:1:10 - error TS1484: 'AppType' is a type and must be imported using a type-only import when 'verbatimModuleSyntax' is enabled.

1 import { AppType } from './server'
           ~~~~~~~


Found 1 error in client.ts:1
```

image

After applying this PR, `tsc --noEmit` exits cleanly with no output.

</details>

Happy to adjust if anything is off.

### DIFF
--- a/docs/concepts/stacks.md
+++ b/docs/concepts/stacks.md
@@ -100,7 +100,7 @@ Then, magically, completion works and the endpoint path and request type are sug
 ![](/images/sc03.gif)
 
 ```ts
-import { AppType } from './server'
+import type { AppType } from './server'
 import { hc } from 'hono/client'
 
 const client = hc<AppType>('/api')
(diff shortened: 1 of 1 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@o8n 

Nice! Thanks.

PR #865, opened 2026-05-25
docs(streaming): use !stream.aborted in SSE example

tim775:
I think `while (!stream.aborted)` is better for the example since `while (true)` keeps the loop going even after the client is gone, leaking work on every disconnect.


yusukebe (approved):
LGTM!

yusukebe:
@tim775 Thanks!

PR #878, opened 2026-06-26
feat(home): replace hero image with a real, highlighted code block

martinezharo:
## Summary

The hero currently shows a static bitmap (`public/images/code.webp`) of the "Hello Hono" snippet. This PR replaces it with the **same snippet rendered as real, syntax‑highlighted text** inside a macOS‑style window.

It's highlighted at **build time with the site's own Shiki themes** (`github-light` / `github-dark`) and frozen as static HTML, so there is no client‑side highlighting.

## Before / After

**Light mode**

| Before | After |
| --- | --- |
| image |image |

**Dark mode**

| Before | After |
| --- | --- |
|image | image |

## Why

- **Light/dark aware** — the old bitmap was dark‑only and looked out of place in light mode (see Before / light). The new block uses `--vp-code-block-bg` and the github‑light/dark themes, exactly like the docs code blocks.
- **Accessibility** — real text: selectable, copyable, scalable with browser zoom, and readable by screen readers, instead of a flat image.
- **Crispness** — vector text stays sharp at any zoom / DPI; no bitmap pixelation.
- **No flash / CLS** — highlighting happens at build time and ships as static HTML, so there is no client‑side highlighting pass and no layout shift. It also drops the `code.webp` image request.
- **Responsive** — the window is sized in container‑query units (`cqw`) with an `aspect-ratio`, so it keeps the original footprint and scales with its width.

## Kawaii mode is unchanged

The `?kawaii=true` easter egg works exactly as before — same image, same size (510×320) and same position. In normal mode you get the code window; in kawaii mode it is swapped for the kawaii logo via CSS. The toggle script in `config.ts` is simplified to only set the class before paint (the swap itself is now handled in CSS).

| Before (kawaii) | After (kawaii) |
| --- | --- |
| image | image |

## Notes

- The snippet is frozen Shiki output; `HeroImage.vue` documents how to regenerate it if the example ever changes.
- `HeroImage.vue` uses `v-html` with that **static, build‑time** HTML (no dynamic or user input → no XSS surface); it is done this way to avoid client‑side highlighting and the flash it causes.

## Verification

bun install --frozen-lockfile
bun run format
bun run build

---

Happy to iterate on the design or close this if it doesn't fit the direction you want for the hero. 🙏


.vitepress/theme/components/HeroImage.vue
yusukebe, line 1: Can we remove `public/images/code.webp` now?

.vitepress/theme/custom.css
yusukebe, line 1: I think the contrast between the code background and the overall background is too subtle. How about making the background of the code darker?

Image

yusukebe:
Hey @martinezharo !

Super cool! I added the comments. Please check!

.vitepress/theme/custom.css
martinezharo: Sure! I've prepared a few proposals. Happy to apply whichever you prefer.

Here are the options for the code block background in light mode:

| Option | Preview |
| --- | --- |
| 1. Darker background (`#eff2f5`)<br>Going much darker starts to hurt the syntax-highlight contrast | Captura de pantalla 2026-06-27 002604 |
| 2. Keep the current color + add a subtle shadow<br>(If you like this one, I can make it light-mode-only or apply it to both modes) | Captura de pantalla 2026-06-27 002142 |
| 3. Use the same appearance in both light and dark mode, closer to the original screenshot | Captura de pantalla 2026-06-27 002814 |

.vitepress/theme/components/HeroImage.vue
martinezharo: Yes, we can. I've already removed it.

.vitepress/theme/custom.css
yusukebe: @martinezharo 

I love "2. Keep the current color + add a subtle shadow"! It's the best. I also want to see the dark-mode pattern.

BlankParticle:
very cool!
i recently just made the fix for flashing kawaii mode, this is good 👍

.vitepress/theme/custom.css
martinezharo: Glad you like option 2! The shadow version is already committed.
Here's how it looks in dark mode:

Captura de pantalla 2026-06-27 094129

---

Also, a heads-up: while making these changes I noticed this PR had
introduced a small regression in kawaii mode. The hero
logo had shifted slightly and changed size compared to the original on
mobile and some other devices. I've fixed it in cab9107436a4ea1cb5b8336bab8502a33d47f8d7, and I've
re-checked that it now matches the original across the widths I tested.

Let me know if anything else needs changing.

yusukebe (approved):
LGTM!

yusukebe:
@martinezharo 

Looks good! This is a great PR in the AI era, when many PRs are sloppy. I'm so glad. Thank you very much!

PR #880, opened 2026-06-29
Update jwt verify docs

RiadhAdrani:
Signing method is [no longer optional](https://github.com/honojs/hono/blob/97c6fe1f12298c715eb7b2da65b4b6e0d81682bb/src/utils/jwt/jwt.ts#L99)

### DIFF
--- a/docs/helpers/jwt.md
+++ b/docs/helpers/jwt.md
@@ -22,7 +22,7 @@ This function generates a JWT token by encoding a payload and signing it using t
 sign(
   payload: unknown,
   secret: string,
-  alg?: 'HS256';
+  alg: 'HS256';
 
 ): Promise<string>;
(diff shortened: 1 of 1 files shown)


docs/helpers/jwt.md
```diff
@@ -22,7 +22,7 @@ This function generates a JWT token by encoding a payload and signing it using t
 sign(
   payload: unknown,
   secret: string,
-  alg?: 'HS256';
+  alg: 'HS256';
```
yusukebe, line 25: We can't accept this. The argument`alg` for `sign` is optional.

Image

RiadhAdrani: updated hono and was looking at verify doc thinking it was sign. closing

PR #886, opened 2026-06-30
docs(middleware): add StitchAPI to third-party middleware list

rejifald:
Adds StitchAPI to the **Utilities** section of the third-party middleware list.

[`@stitchapi/hono`](https://github.com/rejifald/StitchAPI/tree/main/packages/hono) is Hono middleware for [StitchAPI](https://stitchapi.dev): it puts a per-request `seam` on the context (`c.get('stitch')`) — a principal-bound handle for typed, validated, resilient outbound API calls — plus an SSE streaming helper (`streamStitchSse`) and a Stitch-error → HTTP mapper. Every import is from `hono` or `stitchapi` (no `node:*`), so it runs on Node, Cloudflare Workers, Deno, Bun and Vercel Edge.

Placed alphabetically within Utilities (between Session and tsyringe).


yusukebe (approved):
LGTM!

PR #887, opened 2026-07-01
chore: migrate to pnpm

### LINKED ISSUE #365: [Discussion] Changing package manager
following my PR in honojs/middleware, https://github.com/honojs/middleware/pull/2006
I was looking into other hono repos if we should standardize the same package manager across all the repos, it would be weird for the `node-server` to use bun as a runtime in dev/testing, but looks like bun is used for package manager only. 

`honojs/{hono,honox,website}` uses bun, `honojs/{vite-plugins,starter}` uses yarn
Should we put more effort into standardizing all the repos or is it fine to keep these repos on bun/yarn as it is right now?
cc @yusukebe

BlankParticle:
fixes https://github.com/honojs/node-server/issues/365
follow up to https://github.com/honojs/middleware/pull/2006 and https://github.com/honojs/node-server/pull/367


package.json
```diff
@@ -4,7 +4,7 @@
   "type": "module",
   "scripts": {
     "dev": "vitepress dev",
-    "prebuild": "bun ./scripts/build-llm-docs.ts",
+    "prebuild": "node ./scripts/build-llm-docs.ts",
```
yusukebe, line 7: Of course, but it worked on my machine!

yusukebe (approved):
LGTM!

yusukebe:
@BlankParticle 

This also looks good! Thanks!

PR #891, opened 2026-07-21
docs: close app.post call in multiple-validators example

meliharik:
The "multiple validators" example in the validation guide opens `app.post(` but never closes it, so the snippet is unbalanced:

```ts
app.post(
  '/posts/:id',
  validator('param', ...),
  validator('query', ...),
  validator('json', ...),
  (c) => {
    //...
  }
```

The other `app.post(` examples in the same file close properly (line 73 and 275), so this one looks like it just lost its final `)`.


yusukebe (approved):
LGTM!

yusukebe:
@meliharik Thanks!

PR #895, opened 2026-08-09
docs: mention Standard Schema validators in RPC guide

fabian-hiller:
The RPC guide only mentions the Zod Validator. This adds a tip that the Standard Schema Validator works as well, so users can use any Standard Schema library such as Valibot. I'm the author of Valibot.


yusukebe:
@fabian-hiller 

Thank you Fabian!

PR #899, opened 2026-08-20
docs: clarify why graceful shutdown is needed on Node.js but not Bun, Deno and others.

daniel-dia:
## Summary

The Node.js getting-started guide shows how to gracefully shut down the server, but doesn't say *why* this is only needed there and not on the Bun or Deno pages. This adds a short, two-sentence note explaining the reason:

- Node.js's native `http` module predates the Fetch API, so `@hono/node-server` wraps it and exposes the resulting server instance for you to manage.
- Bun and Deno implement the Fetch API natively and own the server lifecycle themselves, so there's no separate server instance for Hono to hand back — which is why those pages don't cover graceful shutdown.

## Test plan

- [x] Docs-only change, no code affected.
- [x] Verified the note renders correctly in the existing Markdown/VitePress structure (plain paragraph, no new components).

Opening as a draft while I get feedback on wording/placement.

### DIFF
--- a/docs/getting-started/nodejs.md
+++ b/docs/getting-started/nodejs.md
@@ -83,8 +83,6 @@ app.get('/', (c) => c.text('Hello Node.js!'))
 serve(app)
 ```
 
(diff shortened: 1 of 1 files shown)


yusukebe (approved):

yusukebe:
@daniel-dia Thanks!

PR #901, opened 2026-08-25
docs: update Scalar theme options for Hono

hanspagel:
The list of themes was out of date. This updates it to the current themes and makes clear that the Hono theme is applied by default.


yusukebe (approved):
LGTM!

yusukebe:
@hanspagel Thanks!

PR #904, opened 2026-09-12
docs(rpc): revert wrong implementation

Zencron:
Revert changes made in https://github.com/honojs/website/pull/727. New implementation does not work, and causes types to be inferred as `any`.

Original implementation from https://github.com/honojs/website/pull/490 should clarify.


yusukebe (approved):
LGTM!

yusukebe:
I couldn't reproduce the `any` issue, but the current code does have a problem. The type isn't precomputed in the `.d.ts`. So reverting is right. Thanks!

PR #908, opened 2026-09-20
fix: add charset=utf-8 to .txt responses via _headers

Debbl:
## Summary

`https://hono.dev/llms.txt` (and `llms-full.txt` / `llms-small.txt`) is served in production with a bare `Content-Type: text/plain`, while `wrangler dev` returns `text/plain; charset=utf-8` locally.

```
$ curl -sI https://hono.dev/llms.txt | grep -i content-type
content-type: text/plain

$ curl -sI http://localhost:8787/llms.txt | grep -i content-type
Content-Type: text/plain; charset=utf-8
```

The difference comes from wrangler: its local asset server appends `charset=utf-8` to every `text/*` type, but the production Static Assets service only maps the extension to `text/plain`. Since these files contain non-ASCII content (🔥, CJK), clients that fall back to ISO-8859-1 show mojibake.

## Fix

Add `public/_headers` with a rule for `*.txt`:

```
/*.txt
  Content-Type: text/plain; charset=utf-8
```

`public/` is copied verbatim into `.vitepress/dist`, so the file ships with the normal build and is picked up by Workers Static Assets.

## Verification

Ran `wrangler dev` against the built `dist` with this file included:

```
[wrangler:info] ✨ Parsed 1 valid header rule.
$ curl -sI http://localhost:8799/llms.txt | grep -i content-type
Content-Type: text/plain; charset=utf-8
```

Note: after deploy, the Cloudflare edge cache (`cf-cache-status: HIT`) may need to expire or be purged before the new header shows up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yusukebe (approved):
LGTM!

yusukebe:
@Debbl 

This fix is meaningful. Thanks!

